PR #81, opened 2017-11-13
:fire: Removes DS_Store

jefftriplett:


simonw:
Hah, I haven't even announced this yet :) Travis is upset because I'm using SQL in the tests which isn't compatible with their version of Python 3.

jefftriplett:
Oops, if I jumped the gun. I saw the project in my github activity feed and saw some low hanging fruit :)

Issue #13, opened 2017-10-23
Add a syntax highlighting SQL editor


simonw:
https://ace.c9.io/#nav=embedding looks like a good option

simonw:
Fixed in https://github.com/simonw/datasette/commit/8252daa4c14d73b4b69e3f2db4576bb39d73c070 - thanks, @tomdyson!

PR #89, opened 2017-11-14
SQL syntax highlighting with CodeMirror

tomdyson:
Addresses #13 

Future enhancements could include autocompletion of table and column names, e.g. with

```javascript
extraKeys: {"Ctrl-Space": "autocomplete"},
hintOptions: {tables: {
      users: ["name", "score", "birthDate"],
      countries: ["name", "population", "size"]
    }}
```

(see https://codemirror.net/doc/manual.html#addon_sql-hint and source at http://codemirror.net/mode/sql/)


simonw:
This is exactly what I was after, thanks!

Issue #57, opened 2017-11-11
Ship a Docker image of the whole thing


simonw:
The generated Docker images can then just inherit from that. This will speed up deploys as no need to `pip install` anything.

- [ ] Ship that image to Docker Hub
- [ ] Update the generated Dockerfile to use it

macropin:
I'm happy to contribute this. Just let me know if you want a Dockerfile for development or production purposes, or both. 

If it's prod then we can just pip install the source from pypi, otherwise for dev we'll need a `requirements.txt` to speed up rebuilds.

macropin:
Let me know if you'd like a PR. The image is usable as 
`docker run --rm -t -i -p 9000:8001 -v $(pwd)/db:/db datasette datasette serve /db/chinook.db`

simonw:
I’m intrigued by this pattern: 

https://github.com/macropin/datasette/commit/147195c2fdfa2b984d8f9fc1c6cab6634970a056#diff-3254677a7917c6c01f55212f86c57fbfR8

What’s the benefit of doing that? Does it result in a smaller image size?

macropin:
The pattern is called "multi-stage builds". And the result is a svelte 226MB image (201MB for 3.6-slim) vs 700MB+ for the full image.

simonw:
Thanks for the explanation! Please do start a pull request.

simonw:
Shipped to Docker Hub: https://hub.docker.com/r/datasetteproject/datasette/

I did this manually the first time. I'll set Travis up to do this automatically in #329

PR #94, opened 2017-11-14
Initial add simple prod ready Dockerfile refs #57

macropin:
Multi-stage build based off official python:3.6-slim

Example usage:
```
docker run --rm -t -i -p 9000:8001 -v $(pwd)/db:/db datasette datasette serve /db/chinook.db
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.6 as build
+
+ARG VERSION=0.11
+RUN pip install datasette==$VERSION
+
+FROM python:3.6-slim
+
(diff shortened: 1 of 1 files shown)


simonw:
Works for me. I'm going to land this.

Just one thing:

    simonw$ docker run --rm -t -i -p 9001:8001 c408e8cfbe40 datasette publish now
     The publish command requires "now" to be installed and configured 
    Follow the instructions at https://zeit.co/now#whats-now

Maybe we should have the Docker container install the "now" client? Not sure how much size that would add though. I think it's OK without for the moment.

Issue #64, opened 2017-11-11
Support for ?field__isnull=1 or similar


simonw:

simonw:
This was fixed by ed2b3f25beac720f14869350baacc5f62b065194 in #107 - thanks @raynae!

PR #107, opened 2017-11-15
add support for ?field__isnull=1

raynae:
Is this what you had in mind for [this issue](https://github.com/simonw/datasette/issues/64)?


simonw:
It is - but I think this will break on this line since it expects two format string parameters:

https://github.com/raynae/datasette/blob/cdaa111e284daa8334bfadd426e1f2c1f2e5503e/datasette/utils.py#L62

Needs unit tests too, which live here: https://github.com/simonw/datasette/blob/f45ca30f91b92ac68adaba893bf034f13ec61ced/tests/test_utils.py#L49

raynae:
Thanks for the guidance.  I added a unit test and made a slight change to utils.py.

I didn't realize this, but evidently string.format only complains if you supply less arguments than there are format placeholders, so the original commit worked, but was adding a superfluous named param.

I added a conditional that prevents the named param from being created and ensures the correct number of args are passed to sting.format. It has the side effect of hiding the SQL query in /templates/table.html when there are no other where clauses--not sure if that's the desired outcome here.

raynae:
Thanks for bearing with me. I was getting a message about my branch diverging when I tried to push after rebasing, so I merged master into isnull, seems like that did the trick. Let me know if I should make any corrections.

PR #114, opened 2017-11-17
Add spatialite, switch to debian and local build

ingenieroariel:
Improves the Dockerfile to support spatial datasets, work with the local datasette code (Friendly with git tags and Dockerhub) and moves to slim debian, a small image easy to extend via apt packages for sqlite.


simonw:
Fantastic! Thank you very much.

Issue #177, opened 2018-01-09
Publishing to Heroku - metadata file not uploaded?


psychemedia:
Trying to run *datasette* (version 0.14)  on Heroku with a `metadata.json` doesn't seem to be picking up the `metadata.json` file? 

On a Mac with dodgy `tar` support:

```
 ▸    Couldn't detect GNU tar. Builds could fail due to decompression errors
 ▸    See
 ▸    https://devcenter.heroku.com/articles/platform-api-deploying-slugs#create-slug-archive
 ▸    Please install it, or specify the '--tar' option
 ▸    Falling back to node's built-in compressor
```

Could that be causing the issue?

Also, I'm not seeing custom query links anywhere obvious when I run the metadata file with a local *datasette* server?

PR #178, opened 2018-01-09
If metadata exists, add it to heroku launch command

psychemedia:
The heroku build does seem to make use of any provided `metadata.json` file.

Add the `--metadata` switch to the Heroku web launch command if a `metadata.json` file is available.

Addresses: https://github.com/simonw/datasette/issues/177


simonw:
Thanks for catching this, merged!

PR #181, opened 2018-01-17
add "format sql" button to query page, uses sql-formatter

### LINKED ISSUE #136: "Reformat SQL" button next to SQL editor textarea
Can use this:

https://github.com/zeroturnaround/sql-formatter
https://zeroturnaround.github.io/sql-formatter/

bsmithgall:
Cool project!

This fixes #136 using the suggested [sql formatter](https://github.com/zeroturnaround/sql-formatter) library. I included the minified version in the bundle and added the relevant scripts to the codemirror includes instead of adding new files, though I could also add new files. I wanted to keep it all together, since the result of the format needs access to the editor in order to properly update the codemirror instance.


simonw:
Here's what this looks like:

2018-04-03 at 8 32 am

I need to figure out the right way to handle licensing of bundled software like this - it's MIT licensed which is compatible with Datasette's Apache 2 license, but I feel like bundled licensed software (including codemirror) needs to be recognized in the README or docs somehow.

simonw:
Let's only show the "Format SQL" button if the user has JavaScript enabled.

simonw:
On the licensing front: it looks like the way Django handles this is to keep the licensing header in the files intact, e.g. https://github.com/django/django/blob/6deaddcca367d0143c815aaa42342021baa3b41e/django/contrib/admin/static/admin/js/vendor/jquery/jquery.js

So for this change, adding a comment at the top of `sql-formatter.min.js` which references the MIT license would do the trick.

bsmithgall:
I can work on that -- would you prefer to inline a `display: hidden` and then have the javascript flip the visibility or include it as css?

simonw:
I'd prefer to have the JavaScript actually manipulate the DOM to add the button - something like this:

    var button = document.createElement('button');
    button.value = 'Format SQL';
    button.addEventListener(
        'click', format, false
    );
    document.getElementById('run-sql').parentNode.appendChild(button);

bsmithgall:
I've implemented that approach in 86ac746. It does cause the button to pop in only after Codemirror is finished rendering which is a bit awkward.

simonw:
Closing this because this feature was shipped in #592

Issue #172, opened 2017-12-11
Show size of .db file next to download link


simonw:
Size in bytes should be calculated by datasette inspect.

Template should display it in KB or MB or GB

simonw:
Demo: https://latest.datasette.io/fixtures-dd88475

<img width="373" alt="screen shot 2019-02-05 at 9 08 49 pm" src="https://user-images.githubusercontent.com/9599/52321564-41f03e00-298a-11e9-9fad-a8bb4e89ad24.png">

PR #182, opened 2018-01-25
Add db filesize next to download link

raynae:
Took a stab at #172, will this do the trick?

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -25,6 +25,7 @@ from .utils import (
     escape_sqlite_table_name,
     filters_should_redirect,
     get_all_foreign_keys,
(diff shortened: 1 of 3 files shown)


datasette/app.py
```diff
@@ -971,6 +973,7 @@ def inspect(self):
                 self._inspect[name] = {
                     'hash': m.hexdigest(),
                     'file': str(path),
+                    'size': get_size(path),
```
simonw: I think the inspect mechanism should store the raw size here as an integer number of bytes. The KB/GB/MB display logic should be handled at display time, maybe as a custom template filter.

PR #200, opened 2018-04-11
Hide Spatialite system tables

russss:
They were getting on my nerves.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -21,6 +21,7 @@ from .utils import (
     CustomJSONEncoder,
     compound_keys_after_sql,
     detect_fts_sql,
+    detect_spatialite,
     escape_css_string,
     escape_sqlite,
     filters_should_redirect,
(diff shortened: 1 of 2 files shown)


simonw:
We should only do this if we're certain the spatialite module has been loaded. I could imagine someone having a `sql_statements_log` table of their own without using spatialite for example.

I think the most reliable way to detect spatialite is to run `SELECT AddGeometryColumn(1, 2, 3, 4, 5);` against a `:memory:` database and see if it throws an exception - similar to how we detect FTS. We could add this as a `detect_spatialite()` function in `utils.py` and call it once on startup.

datasette/app.py
```diff
@@ -1121,7 +1121,7 @@ def inspect(self):
                         tables[table]['foreign_keys'] = info
 
                     # Mark tables 'hidden' if they relate to FTS virtual tables
```
simonw: Need to update this comment

russss:
> I think the most reliable way to detect spatialite is to run `SELECT AddGeometryColumn(1, 2, 3, 4, 5);` against a `:memory:` database and see if it throws an exception

Or just see if there's a `geometry_columns` table? I think that's quite unlikely to be added by accident (and it's an OGC standard). It also tells you if Spatialite is installed in the database rather than just loaded.

simonw:
Nice, thanks very much.

Issue #203, opened 2018-04-12
Support for units


russss:
It would be nice to be able to attach a unit to a column in the metadata, and have it rendered with that unit (and SI prefix) when it's displayed.

It would also be nice to support entering the prefixes in variables when querying.

With my radio licensing app I've put all frequencies in Hz. It's easy enough to special-case the row rendering to add the SI prefixes, but it's pretty unusable when querying by that field.

simonw:
I like this. I'd like to be able to attach a full description to a column as well. We could support these in `metadata.json`

simonw:
This also feeds into the visualization features I want to add - we could use this kind of metadata to automatically apply meaningful labels to graphs.

russss:
Looks like [pint](https://pint.readthedocs.io/en/latest/tutorial.html) is pretty good at this.

```python
In [1]: import pint

In [2]: ureg = pint.UnitRegistry()

In [3]: q = 3e6 * ureg('Hz')

In [4]: '{:~P}'.format(q.to_compact())
Out[4]: '3.0 MHz'

In [5]: q = 0.3 * ureg('m')

In [5]: '{:~P}'.format(q.to_compact())
Out[5]: '300.0 mm'

In [6]: q = 5 * ureg('')

In [7]: '{:~P}'.format(q.to_compact())
Out[7]: '5'
```

simonw:
This is really cool - I'm very impressed by pint.

I'd like to figure out a sensible opt-in way to expose this in the JSON output as well. Maybe with a `&_units=true` parameter?

We should definitely expose the units section from the table metadata in the output of https://wtr-api.herokuapp.com/wtr-663ea99/license_frequency.json

simonw:
In #204 you said "I'd like to add support for using units when querying but this is PR is pretty usable as-is." - I'm fascinated to hear more about how this could work.

russss:
> I'd like to figure out a sensible opt-in way to expose this in the JSON output as well. Maybe with a &_units=true parameter?

From a machine-readable perspective I'm not sure why it would be useful to decorate the values with units. The unit metadata should definitely be exposed in the JSON, though.

> In #204 you said "I'd like to add support for using units when querying but this is PR is pretty usable as-is." - I'm fascinated to hear more about how this could work.

I'm thinking about a couple of approaches here. I think the simplest one is: if the column has a unit attached, optionally accept units in query fields:

```python
column_units = ureg("Hz") #  Create a unit object for the column's unit
[...]

PR #205, opened 2018-04-14
Support filtering with units and more

russss:
The first commit:
* Adds units to exported JSON
* Adds units key to metadata skeleton
* Adds some docs for units

The second commit adds filtering by units by the first method I mentioned in #203:
image

I think it integrates pretty neatly.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -647,7 +647,9 @@ class TableView(RowTableShared):
                 forward_querystring=False
             )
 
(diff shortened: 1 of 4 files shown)


simonw:
This looks great so far - love the new documentation. Let's throw in a unit test or two for the basic unit filters (mainly as a protection against accidental regressions in the future).

russss:
I've added some tests and that docs link.

PR #207, opened 2018-04-14
Link foreign keys which don't have labels

russss:
This renders unlabeled FKs as simple links. I can't see why this would cause any major problems.

image

Also includes bonus fixes for two minor issues:

* In foreign key link hrefs the primary key was escaped using HTML escaping rather than URL escaping. This broke some non-integer PKs.
* Print tracebacks to console when handling 500 errors.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -17,6 +17,7 @@ import jinja2
 import hashlib
 import time
 import pint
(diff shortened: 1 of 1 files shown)


simonw:
I'm going to merge this and then add a unit test.

PR #209, opened 2018-04-15
 Don't duplicate simple primary keys in the link column

russss:
When there's a simple (single-column) primary key, it looks weird to duplicate it in the link column.

This change removes the second PK column and treats the link column as if it were the PK column from a header/sorting perspective. 

This might make it a bit more difficult to tell what the link for the row is, I'm not sure yet. I feel like the alternative is to change the link column to just have the text "view" or something, instead of repeating the PK. (I doubt it makes much more sense with compound PKs.)

Bonus change in this PR: fix urlencoding of links in the displayed HTML.

Before:
image

After:
image

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -524,10 +524,11 @@ class RowTableShared(BaseView):
                 cells.append({
                     'column': 'Link',
                     'value': jinja2.Markup(
-                        '<a href="/{database}/{table}/{flat_pks}">{flat_pks}</a>'.format(
+                        '<a href="/{database}/{table}/{flat_pks_quoted}">{flat_pks}</a>'.format(
                             database=database,
                             table=urllib.parse.quote_plus(table),
-                            flat_pks=path_from_row_pks(row, pks, not pks),
+                            flat_pks=str(jinja2.escape(path_from_row_pks(row, pks, not pks, False))),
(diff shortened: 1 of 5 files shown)


russss:
I suspected this would cause some test failures, but I'll wait for opinions before attempting to fix them.

simonw:
I think this is a good improvement. If you fix the tests I'll merge it.

russss:
Tests now fixed, honest. The failing test on Travis looks like an intermittent sqlite failure which should resolve itself on a retry...

russss:
I've added another commit which puts classes a class on each `<td>` by default with its column name, and I've also made the PK column bold.

Unfortunately the tests are still failing on 3.6, which is weird. I can't reproduce locally...

simonw:
I managed to get a better error message out of that text. The server is returning this (but only on Python 3.6, not on Python 3.5 - and only in Travis, not in my local environment):

```{'error': 'interrupted', 'ok': False, 'status': 400, 'title': 'Invalid SQL'}```

https://travis-ci.org/simonw/datasette/jobs/367929134

simonw:
OK, aaf59db570ab7688af72c08bb5bc1edc145e3e07 should mean that the tests pass when I merge that.

PR #258, opened 2018-05-14
Add new metadata key persistent_urls which removes the hash from all database urls

philroche:
Add new metadata key "persistent_urls" which removes the hash from all database urls when set to "true"

This PR is just to gauge if this, or something like it, is something you would consider merging?

I understand the reason why the substring of the hash is included in the url but
there are some use cases where the urls should persist across deployments. For bookmarks
for example or for scripts that use the JSON API.

This is the initial commit for this feature. Tests and documentation updates to follow.

### DIFF
--- a/datasette/cli.py
+++ b/datasette/cli.py
@@ -255,6 +255,7 @@ def skeleton(files, metadata, sqlite_extensions):
                 "license_url": None,
                 "source": None,
                 "source_url": None,
+                "persistent_urls": "false",
                 "databases": databases,
             },
             indent=4,
--- a/datasette/templates/database.html
+++ b/datasette/templates/database.html
@@ -16,7 +16,7 @@
 
 {% block description_source_license %}{% include "_description_source_license.html" %}{% endblock %}
 
-<form class="sql" action="/{{ database }}-{{ database_hash }}" method="get">
+<form class="sql" action="/{{ database_url }}" method="get">
     <h3>Custom SQL query</h3>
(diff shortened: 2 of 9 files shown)


simonw:
The URL does persist across deployments already, in that you can use the URL without the hash and it will redirect to the current location. Here's an example of that: https://san-francisco.datasettes.com/sf-trees/Street_Tree_List.json

This also works if you attempt to hit the incorrect hash, e.g. if you have deployed a new version of the database with an updated hash. The old hash will redirect, e.g. https://san-francisco.datasettes.com/sf-trees-c4b972c/Street_Tree_List.json

If you serve Datasette from a HTTP/2 proxy (I've been using Cloudflare for this) you won't even have to pay the cost of the redirect - Datasette sends a `Link: <URL>; rel=preload` header with those redirects, which causes Cloudflare to push out the redirected source as part of that HTTP/2 request. You can fire up the Chrome DevTools to watch this happen.

https://github.com/simonw/datasette/blob/2b79f2bdeb1efa86e0756e741292d625f91cb93d/datasette/views/base.py#L91

All of that said... I'm not at all opposed to this feature. For consistency with other Datasette options (e.g. `--cors`) I'd prefer to do this as an optional argument to the `datasette serve` command - something like this:

    datasette serve mydb.db --no-url-hash

philroche:
Excellent, I was not aware of the auto redirect to the new hash. My bad

This solves my use case.

I do agree that your suggested --no-url-hash approach is much neater. I will investigate

Issue #276, opened 2018-05-21
Handle spatialite geometry columns better


russss:
I'd like to see spatialite geometry columns rendered more sensibly - at the moment they come through as well-known-binary unless you use custom SQL, and WKB isn't of much use to anyone on the web.

In HTML: they should be shown either as simple lat/long (if it's just a point, for example), or as a sensible placeholder if they're more complex geometries.

In JSON: they should be GeoJSON geometries, (which means they can be automatically fed into a leaflet map with no further messing around).

In CSV: they should be WKT.

I briefly wondered if this should go into a plugin, but I suspect it needs hooking in at a deeper level than the plugin architecture will support any time soon.

simonw:
This probably needs to be in a plugin simply because getting Spatialite compiled and installed is a bit of a pain.

It's a great opportunity to expand the plugin hooks in useful ways though.

russss:
Well, we do have the capability to detect spatialite so my intention certainly wasn't to require it. 

I can see the advantage of having it as a plugin but it does touch a number of points in the code. I think I'm going to attack this by refactoring the necessary bits and seeing where that leads (which was my plan anyway).

I think my main concern is - if I add certain plugin hooks for this, is anything else ever going to use them? I'm not sure I have an answer to that question yet, either way.

simonw:
Right now the plugin stuff is early enough that I'd like to get as many potential plugin hooks as possible crafted out A much easier to judge if they should be added as actual hooks if we have a working branch prototype of them.

Some kind of mechanism for custom column display is already needed - eg there are columns where I want to say "render this as markdown" or "URLify any links in this text" - or even "use this date format" or "add commas to this integer".

You can do it with a custom template but a lower-level mechanism would be nicer.

russss:
Yup, I'll have a think about it. My current thoughts are for spatialite we'll need to hook into the following places:

* Inspection, so we can detect which columns are geometry columns. (We also currently ignore spatialite tables during inspection, it may be worth moving that to the plugin as well.)
* After data load, so we can convert WKB into the correct intermediate format for display. The alternative here is to alter the select SQL itself and get spatialite to do this conversion, but that strikes me as a bit more complex and possibly not as useful.
* HTML rendering.
* Querying?

The rendering and querying hooks could also potentially be used to move the units support into a plugin.

simonw:
[...]

PR #277, opened 2018-05-21
Refactor inspect logic

russss:
This pulls the logic for inspect out into a new file which makes it a bit easier to understand.

This was going to be the first part of an implementation for #276, but it seems like that might take a while so I'm going to PR a few bits of refactoring individually.


simonw:
This is definitely a big improvement.

I'd like to refactor the unit tests that cover .inspect() too - currently they are a huge ugly blob at the top of test_api.py

simonw:
We should merge his before refactoring the tests though, because that way we don't couple the new tests to the verification of this change.

PR #279, opened 2018-05-22
Add version number support with Versioneer

### LINKED ISSUE #273: Figure out a way to have /-/version return current git commit hash
https://fivethirtyeight.datasettes.com/-/versions reports Datasette version `0.21`

This isn't actually correct. The deploy script for that site actually deploys current master using `https://github.com/simonw/datasette/archive/master.zip`: https://github.com/simonw/fivethirtyeight-datasette/blob/66b4b0dfedd7237bc8c02d3e26d905bca7b84069/Dockerfile#L9

Ideally this would show the current commit hash, but I'm not at all sure if it's possible to derive that from `pip install https://github.com/simonw/datasette/archive/master.zip`. Is there another mechanism that could be used to reliably `pip install` current master but still provide access to the most recent commit hash?

rgieseke:
I think that's all for getting Versioneer support, I've been happily using it in a couple of projects ... 

```
In [2]: datasette.__version__
Out[2]: '0.22+3.g6e12445'
```
Repo:
https://github.com/warner/python-versioneer

Versioneer Licence:
Public Domain (CC0-1.0)

Closes #273


simonw:
This is fantastic!

I think I prefer the aesthetics of just "0.22" for the version string if it's a tagged release with no additional changes - does that work?

I'd like to continue to provide a tuple that can be imported from the version.py module as well, as seen here:

https://github.com/simonw/datasette/blob/558d9d7bfef3dd633eb16389281b67d42c9bdeef/datasette/version.py#L1

Presumably we can generate that from the versioneer string?

rgieseke:
Alright, that should work now -- let me know if you would prefer any different behaviour.

Issue #278, opened 2018-05-22
Build smallest possible Docker image with Datasette plus recent SQLite (with json1) plus Spatialite 4.4.0


simonw:
A Dockerfile that does the following:

* Bundles Datasette master
* Python 3.6 most recent version (or 3.7 if it has been released)
* SQLite 3.23.1 (or most recent release) such that "import sqite3" in Python gets that version. Ideally with the json1 module baked in by default, but having it loadable as an optional module is fine too
* SpatiaLite 4.4.0-RC0 (or most recent version) such that it can be loaded as an optional module
* Uses multi-stage builds to stay as small as possible

Note that the current "release" of SpatiaLite is 4.3.0 which is missing key features like https://www.gaia-gis.it/fossil/libspatialite/wiki?name=KNN - 4.4.0 probably needs to be compiled from source.

I don't know the best way to get a current SQLite version bundled for Python 3. Maybe https://github.com/coleifer/pysqlite3 ?

simonw:
For SpatiaLite this example may be useful - though it's building 4.3.0 and not 4.4.0: https://github.com/terranodo/spatialite-docker/blob/master/Dockerfile

PR #280, opened 2018-05-22
Build Dockerfile with recent Sqlite + Spatialite

r4vi:
This solves #278 without bloating the Dockerfile too much, the image size is now
495MB (original was ~240MB) but it could be reduced significantly if we only
copied the output of the compilation of spatialite and friends to
/usr/local/lib, instead of the entirety of it however that will take more time.

In the python code change references to `import sqlite3` to `import pysqlite3`
and it should use the compiled version of sqlite3.23.1. You don't need to
try/except because pysqlite3 falls back to builtin sqlite3 if there is no
compiled version.

```bash
  $ docker run --rm -it datasette spatialite
  SpatiaLite version ..: 4.4.0-RC0	Supported Extensions:
    - 'VirtualShape'	[direct Shapefile access]
    - 'VirtualDbf'		[direct DBF access]
    - 'VirtualXL'		[direct XLS access]
    - 'VirtualText'		[direct CSV/TXT access]
    - 'VirtualNetwork'	[Dijkstra shortest path]
    - 'RTree'		[Spatial Index - R*Tree]
    - 'MbrCache'		[Spatial Index - MBR cache]
    - 'VirtualSpatialIndex'	[R*Tree metahandler]
    - 'VirtualElementary'	[ElemGeoms metahandler]
    - 'VirtualKNN'	[K-Nearest Neighbors metahandler]
    - 'VirtualXPath'	[XML Path Language - XPath]
    - 'VirtualFDO'		[FDO-OGR interoperability]
    - 'VirtualGPKG'	[OGC GeoPackage interoperability]
    - 'VirtualBBox'		[BoundingBox tables]
    - 'SpatiaLite'		[Spatial SQL - OGC]
  PROJ.4 version ......: Rel. 4.9.3, 15 August 2016
  GEOS version ........: 3.5.1-CAPI-1.9.1 r4246
  TARGET CPU ..........: x86_64-linux-gnu
  the SPATIAL_REF_SYS table already contains some row(s)
  SQLite version ......: 3.23.1
  Enter ".help" for instructions
  SQLite version 3.23.1 2018-04-10 17:39:29
  Enter ".help" for instructions
  Enter SQL statements terminated with a ";"
  spatialite>
```

```bash
$ docker run --rm -it datasette python -c "import pysqlite3; print(pysqlite3.sqlite_version)"
3.23.1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,41 @@
 FROM python:3.6-slim-stretch as build
 
 # Setup build dependencies
-RUN apt update
-RUN apt install -y python3-dev gcc libsqlite3-mod-spatialite
+RUN apt update \
+&& apt install -y python3-dev build-essential wget libxml2 libxml2-dev libproj-dev libgeos-dev libsqlite3-dev zlib1g-dev pkg-config unzip \
+ && apt clean
+
+
+RUN wget "https://www.sqlite.org/2018/sqlite-autoconf-3230100.tar.gz" && tar xzvf sqlite-autoconf-3230100.tar.gz \
+    && cd sqlite-autoconf-3230100 && ./configure --disable-static --enable-fts5 --enable-json1 CFLAGS="-g -O2 -DSQLITE_ENABLE_FTS3=1 -DSQLITE_ENABLE_FTS4=1 -DSQLITE_ENABLE_RTREE=1" \
+    && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/freexl-1.0.5.tar.gz" && tar zxvf freexl-1.0.5.tar.gz && cd freexl-1.0.5 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/libspatialite-4.4.0-RC0.tar.gz" && tar zxvf libspatialite-4.4.0-RC0.tar.gz
+RUN cd libspatialite-4.4.0-RC0 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/readosm-1.1.0.tar.gz" && tar zxvf readosm-1.1.0.tar.gz && cd readosm-1.1.0 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/spatialite-tools-4.4.0-RC0.tar.gz" && tar zxvf spatialite-tools-4.4.0-RC0.tar.gz && cd spatialite-tools-4.4.0-RC0 && ./configure && make && make install
+
+
 # Add local code to the image instead of fetching from pypi.
 ADD . /datasette
 
 RUN pip install /datasette
 
+
 FROM python:3.6-slim-stretch
 
 # Copy python dependencies
-COPY --from=build /usr/local/lib/python3.6/site-packages /usr/local/lib/python3.6/site-packages
+COPY --from=build /usr/local/lib/ /usr/local/lib/
 # Copy executables
 COPY --from=build /usr/local/bin /usr/local/bin
 # Copy spatial extensions
 COPY --from=build /usr/lib/x86_64-linux-gnu /usr/lib/x86_64-linux-gnu
+
+ENV LD_LIBRARY_PATH=/usr/local/lib
 
 EXPOSE 8001
 CMD ["datasette"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,33 +2,34 @@ FROM python:3.6-slim-stretch as build
 
 # Setup build dependencies
 RUN apt update \
-&& apt install -y python3-dev build-essential wget libxml2 libxml2-dev libproj-dev libgeos-dev libsqlite3-dev zlib1g-dev pkg-config unzip \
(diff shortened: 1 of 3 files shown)


r4vi:
```python
>>> import sqlite3
>>> sqlite3.sqlite_version
'3.23.1'
>>> 
```
running the above in the container seems to show 3.23.1 too so maybe we don't need pysqlite3 at all?

simonw:
This looks amazing!  Can't wait to try this out this evening.

simonw:
Yeah let's try this without pysqlite3 and see if we still get the correct version.

r4vi:
I'm going to clean this up for consistency tomorrow morning so hold off
merging until then please

Dockerfile
```diff
@@ -1,21 +1,41 @@
...
+RUN wget "https://www.gaia-gis.it/gaia-sins/libspatialite-4.4.0-RC0.tar.gz" && tar zxvf libspatialite-4.4.0-RC0.tar.gz
+RUN cd libspatialite-4.4.0-RC0 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/readosm-1.1.0.tar.gz" && tar zxvf readosm-1.1.0.tar.gz && cd readosm-1.1.0 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/spatialite-tools-4.4.0-RC0.tar.gz" && tar zxvf spatialite-tools-4.4.0-RC0.tar.gz && cd spatialite-tools-4.4.0-RC0 && ./configure && make && make install
+
+
 # Add local code to the image instead of fetching from pypi.
 ADD . /datasette
```
larsclaussen: Docker best practices: 

> Although ADD and COPY are functionally similar, generally speaking, COPY is preferred. That’s because it’s more transparent than ADD

Dockerfile
```diff
@@ -1,21 +1,41 @@
 FROM python:3.6-slim-stretch as build
 
 # Setup build dependencies
-RUN apt update
-RUN apt install -y python3-dev gcc libsqlite3-mod-spatialite
+RUN apt update \
+&& apt install -y python3-dev build-essential wget libxml2 libxml2-dev libproj-dev libgeos-dev libsqlite3-dev zlib1g-dev pkg-config unzip \
+ && apt clean
```
larsclaussen: maybe ``&& rm -rf /var/lib/apt/lists/* /tmp/* `` too?

Dockerfile
```diff
@@ -1,21 +1,41 @@
...
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/freexl-1.0.5.tar.gz" && tar zxvf freexl-1.0.5.tar.gz && cd freexl-1.0.5 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/libspatialite-4.4.0-RC0.tar.gz" && tar zxvf libspatialite-4.4.0-RC0.tar.gz
+RUN cd libspatialite-4.4.0-RC0 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/readosm-1.1.0.tar.gz" && tar zxvf readosm-1.1.0.tar.gz && cd readosm-1.1.0 && ./configure && make && make install
+
+RUN wget "https://www.gaia-gis.it/gaia-sins/spatialite-tools-4.4.0-RC0.tar.gz" && tar zxvf spatialite-tools-4.4.0-RC0.tar.gz && cd spatialite-tools-4.4.0-RC0 && ./configure && make && make install
+
```
larsclaussen: are the tar files removed after the installation? That will make the image smaller too

r4vi: @larsclaussen they aren't but they're part of the builder image not the app image so it shouldn't make a difference (same for `&& rm -rf /var/lib/apt/lists/* /tmp/*`) - unless I'm not understanding the builder image mechanism correctly.

larsclaussen: Ah..didn't catch the multi stage build. You're right, shouldn't make a difference then!

r4vi:
Running:
```bash
docker run -p 8001:8001 -v `pwd`:/mnt datasette \
    datasette -p 8001 -h 0.0.0.0 /mnt/fixtures.db \
        --load-extension=/usr/local/lib/mod_spatialite.so
```

is now returning FTS5 enabled in the versions output:

```json
{
    "datasette": {
        "version": "0.22"
    },
    "python": {
        "full": "3.6.5 (default, May  5 2018, 03:07:21) \n[GCC 6.3.0 20170516]",
        "version": "3.6.5"
    },
    "sqlite": {
        "extensions": {
            "json1": null,
            "spatialite": "4.4.0-RC0"
        },
        "fts_versions": [
            "FTS5",
            "FTS4",
            "FTS3"
        ],
        "version": "3.23.1"
    }
}
```
The old query didn't work because specifying `(t TEXT)` caused an error

simonw:
@r4vi any objections to me merging this?

r4vi:
It's good to go @simonw

simonw:
Thank you very much! This is most excellent.

PR #281, opened 2018-05-23
Reduces image size using Alpine + Multistage (re: #278)

iMerica:
Hey Simon! 

I got the image size down from 256MB to 110MB. 

Seems to be working okay, but you might want to test it a bit more.

Example output of `docker run --rm -it <my-tag> datasette`
```
Serve! files=() on port 8001
[2018-05-23 05:23:08 +0000] [1] [INFO] Goin' Fast @ http://127.0.0.1:8001
[2018-05-23 05:23:08 +0000] [1] [INFO] Starting worker [1]
```

Related: https://github.com/simonw/datasette/issues/278

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,22 @@
-FROM python:3.6-slim-stretch as build
-
+FROM python:3.6-alpine as base
+FROM base as builder
 # Setup build dependencies
-RUN apt update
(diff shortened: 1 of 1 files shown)


simonw:
I'm afraid I just merged #280 which means this no longer applies. You're very welcome to see if you can further optimize the new Dockerfile though.

PR #293, opened 2018-05-28
Support for external database connectors

jsancho-gpl:
I think it would be nice that Datasette could work with other file formats that aren't SQLite, like files with PyTables format. I've tried to accomplish that using external connectors published with entry points.

These external connectors must have a structure similar to the structure [PyTables Datasette connector](https://github.com/PyTables/datasette-pytables) has.


jsancho-gpl:
I close this PR because it's better to use the new one #364

PR #345, opened 2018-07-14
Allow app names for `datasette publish heroku`

russss:
Lets you supply the `-n` parameter for Heroku deploys, which also lets you update existing Heroku deployments.


simonw:
Fantastic, we really needed this.

PR #363, opened 2018-08-29
Search all apps during heroku publish

kevboh:
Adds the `-A` option to include apps from all organizations when searching app names for publish.


kevboh:
It looks like the check passed, not sure why it's showing as running in GH.

PR #367, opened 2018-10-27
Mark codemirror files as vendored

jaap3:
GitHub lists datasette as a Javascript project, primarily because of the vendored codemirror files. This is somewhat confusing when you're looking for datasette, knowing it's written in Python.

Luckily it's possible exclude certain files from GitHub's code statistics: https://github.com/github/linguist#using-gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 datasette/_version.py export-subst
+datasette/static/codemirror-* linguist-vendored


simonw:
This is a neat fix, thanks!

simonw:
Before applying this fix, GitHub showed the following statistics:

    Python 50.1% JavaScript 46.0% HTML 3.0% Other 0.9% 

Afterwards, it shows:



    Python 92.8% HTML 5.5% CSS 1.3% Dockerfile 0.4%

PR #390, opened 2018-12-01
tiny typo in customization docs

jaywgraves:
was looking to add some custom templates to my use of datasette and saw this small typo.


simonw:
Thanks for spotting this!

PR #424, opened 2019-03-31
Column types in inspected metadata

russss:
This PR does two things:

* Adds the sqlite column type for each column to the inspected table info.
* Stops binary columns from being rendered to HTML, unless a plugin handles it.

There's a bit more detail in the changeset descriptions.

These changes are intended as a precursor to a plugin which adds first-class support for Spatialite geographic primitives, and perhaps more useful geo-stuff.


russss:
This is pretty conflicty because I forgot how to use git fetch. If you're interested in merging this I'll rewrite it against an actual modern checkout...

russss:
Actually no, I ended up not using the inspected column types in my plugin, and the binary column issue can be solved a lot more simply, so I'll close this.

PR #426, opened 2019-04-10
Upgrade to Jinja2==2.10.1

simonw:
https://nvd.nist.gov/vuln/detail/CVE-2019-10906

This is only a security issue of concern if evaluating templates from untrusted sources, which isn't something I would ever expect a Datasette user to do.


carlmjohnson:
Can you cut a new release with this?

Issue #427, opened 2019-04-11
New design for facet abstraction, including querystring and metadata.json


simonw:
I need a better design for query strings for facets (and for how facets are enabled in `metadata.json`).

Think of all of the potential kinds of facets:

* `?_facet_array=tags` where tags is a JSON array of values
* `_facet_date=datetimecol` - faceted by date part of a datetime
* `_facet_bins=numeric_column` - can I do some kind of fancy binning here? Might need to take an argument
* `?_facet_bins=numeric_column:5` - could be a way to take an argument. We’ll ignore columns with a : in their name.
* `?_facet_json=jsoncol:jsonpath` - could use a JSON path to extract out something to facet on?
* `?_facet_percentile=numericcolumn` - could this work?
* `?_facet_function=column:sqlfunctionname` - maybe this could be interesting? Would allow for e.g. facet by soundex
* `?_facet_prefix=column:prefix` - facet by terms but only if they start with a specific prefix
* `?_facet_substring=column:3,6` - facet by a substr(column, 3, 6)

Maybe bundling JSON in querystrings is a way to do options?

`?_facet_distance={"latitude_column":"x",...}`

Could detect values starting with `{` - and if for some weird reason you have a column starting with that character you can pass this instead: `?_facet_percentile={"column": "{value}"}`

This could even be the mechanism that allows us to extend regular facets to support additional options like adding a sum or max to each one.

Problem: it’s not obvious what the name associated with these facets should be. What if one column is faceted multiple times using multiple facet variants?

Maybe just number them? name1=… name2=… etc?

Other option is to use Solr style querystring syntax for notation. Solr does this: `?f.price.facet.range.gap=100&f.age.facet.range.gap=10`

So how about this:

`?_facet_range=age&_facet_range.span=5`

Related: #359

simonw:
I quite like the Solr idea. It could look like this for Datasette:

`?_facet=name` - default behaviour, same as today. But that's actually an alias for `?_facet.name=name` - which defines a name for the facet.

`?_facet.tags.array=tags` - would define a facet called `tags` that uses an array facet against the `tags` column.

I don't like the need to say `tags` twice in that though.

simonw:
OK, I have a plan:

`?_facet=foo`
`?_facet_facettype=options`

Options here can be one of the following:

- A single value which is the name of a table
- A comma separated list of options
- A JSON object starting with `{` or `[`

[...]

PR #432, opened 2019-04-13
Refactor facets to a class and new plugin, refs #427

simonw:
WIP for #427


simonw:
This has a bug which isn't being caught by the unit tests (yet) - facet suggestion suggests facets that have already been enabled:

fixtures__facetable__15_rows

There is also a test failure due to missing plugin hook documentation.

simonw:
It would be nice to decouple the `request` object from the `Facet` class. The request is needed for two things at the moment:

* To decide if a specific facet bucket has been selected or not
* To construct the `toggle_url` for turning the selection on or off

Can I pull those needs out of the Facet class somehow?

russss:
> Can I pull those needs out of the Facet class somehow?

I was thinking that it might be handy for datasette to have a request object which wraps the Sanic Request. This could include the datasette-specific querystring decoding and the `special_args` parsing from TableView.data.

This would mean that we could expose the request object to plugin hooks without coupling them to Sanic.

simonw:
Abandoning this in favour of #445 - which contains the code from this branch but updated to incorporate recent changes in master.

Issue #400, opened 2019-01-12
datasette publish cloudrun plugin


rprimet:
Google announced that they may launch a simple service for running Docker containers (previously serverless containers, now called "cloud run" -- link to alpha [here](https://services.google.com/fb/forms/serverlesscontainers/)). If/when this happens, it might be a good fit for publishing datasettes? (at least using the current version, manually publishing a datasette seems relatively painless).

simonw:
I'm really excited about this - it looks like it could be a great plugin.

PR #434, opened 2019-04-17
[WIP] publish on google cloud run

rprimet:
This is a very rough draft to start a discussion on a possible datasette cloud run publish plugin (see issue #400).

The main change was to dynamically set the listening port in `make_dockerfile` to satisfy cloud run's [requirements](https://cloud.google.com/run/docs/reference/container-contract).

This was done by running `datasette` through `sh` to get environment variable substitution. Not sure if that's the right approach?

### DIFF
--- a/datasette/publish/cloudrun.py
+++ b/datasette/publish/cloudrun.py
@@ -1,0 +1,71 @@
+from datasette import hookimpl
+import click
+import json
+from subprocess import check_call, check_output
+
+from .common import (
+    add_common_publish_arguments_and_options,
+    fail_if_publish_binary_not_installed,
+)
+from ..utils import temporary_docker_directory
+
+
+@hookimpl
+def publish_subcommand(publish):
+    @publish.command()
+    @add_common_publish_arguments_and_options
+    @click.option(
+        "-n",
+        "--name",
+        default="datasette",
+        help="Application name to use when deploying",
+    )
+    @click.option("--spatialite", is_flag=True, help="Enable SpatialLite extension")
+    def cloudrun(
+        files,
+        metadata,
+        extra_options,
+        branch,
+        template_dir,
+        plugins_dir,
+        static,
+        install,
+        version_note,
+        title,
+        license,
+        license_url,
+        source,
+        source_url,
+        about,
+        about_url,
+        name,
+        spatialite,
+    ):
+        fail_if_publish_binary_not_installed("gcloud", "Google Cloud", "https://cloud.google.com/sdk/")
+        project = check_output("gcloud config get-value project", shell=True).decode("utf-8").strip()
+
+        with temporary_docker_directory(
+            files,
+            name,
+            metadata,
+            extra_options,
+            branch,
+            template_dir,
+            plugins_dir,
+            static,
+            install,
+            spatialite,
+            version_note,
+            {
+                "title": title,
+                "license": license,
+                "license_url": license_url,
+                "source": source,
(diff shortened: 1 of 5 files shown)


simonw:
Thanks for looking into this!

To clarify: currently, the Dockerfile that we generate looks something like this:

```
CMD ["datasette", "serve", "--host", "0.0.0.0", "fixtures.db", "--cors", "--port", "8001"]
```

Your code here changes that CMD line to look like this instead, in order to set the port based on an environment variable:

```
CMD ["sh", "-c", "datasette serve --port $PORT ..."]
```

I wonder if this is the only way to do this?

datasette/publish/cloudrun.py
```diff
@@ -0,0 +1,71 @@
...
+                "title": title,
+                "license": license,
+                "license_url": license_url,
+                "source": source,
+                "source_url": source_url,
+                "about": about,
+                "about_url": about_url,
+            },
+        ):
+            image_id = f"gcr.io/{project}/{name}"
```
simonw: Datasette still supports Python 3.5 so we can't use format strings just yet.

simonw:
I asked @andrewgodwin about this and he confirmed that if we want to read an environment variable we can't use the `CMD [...]` syntax in the way that we were using it.

He did suggest that if we're doing `CMD ["sh", "-c", "datasette serve --port $PORT ..."]` we may as well do this instead:

`CMD "datasette serve --port $PORT ..."`

We should apply some command-line escaping here - if the user passes `--version-note=hello$there` to `datasette publish` we need that $ not to be accidentally evaluated as an environment variable.

It looks like [shlex.quote](https://docs.python.org/dev/library/shlex.html#shlex.quote) is the right way to do that.

simonw:
This is amazing - works an absolute treat. Thank you very much!

eyeseast:
This is exactly what I needed. Thank you.

simonw:
Here's my first working deployment: https://datasette-j7hipcg4aq-uc.a.run.app/fixtures-c35b6a5/facetable?_facet_array=tags

I deployed it using this:

    datasette publish cloudrun fixtures.db --branch=master

The second time I ran the command I got an error:

    ERROR: (gcloud.beta.run.deploy) Deployment endpoint was not found. Perhaps the
    provided region was invalid. Set the `run/region` property to a valid region and
    retry. Ex: `gcloud config set run/region us-central1`

So I ran the command it suggested and then everything worked:

    gcloud config set run/region us-central1
    datasette publish cloudrun fixtures.db --branch=master

simonw:
Documentation is now available here: https://datasette.readthedocs.io/en/latest/publish.html#publishing-to-google-cloud-run

rprimet:
> The second time I ran the command I got an error:

ERROR: (gcloud.beta.run.deploy) Deployment endpoint was not found. Perhaps the
provided region was invalid. Set the `run/region` property to a valid region and
retry. Ex: `gcloud config set run/region us-central1`

Yes, I was able to reproduce this; I used to get prompted for a run region interactively by the `gcloud` tool before, but maybe this is changing? (the [documentation](https://cloud.google.com/run/docs/deploying) now assumes `run/region` is set).

Not sure which course of action is best: making `datasette` ensure that `run/region` is set beforehand or wait a bit until the gcloud CLI stabilizes?

simonw:
Since there's a useful error message I'm OK with revisiting this in a few weeks to see if they change the CLI tool.

PR #439, opened 2019-04-29
[WIP] Add primary key to the extra_body_script hook arguments

russss:
This allows the row to be identified on row pages. The context here is that I want to access the row's data to plot it on a map.

I considered passing the entire template context through to the hook function. This would expose the actual row data and potentially avoid a further fetch request in JS, but it does make the plugin API a lot more leaky. 

(At any rate, using the selected row data is tricky in my case because of Spatialite's infuriating custom binary representation...)

### DIFF
--- a/datasette/hookspecs.py
+++ b/datasette/hookspecs.py
@@ -26,7 +26,7 @@ def extra_js_urls(template, database, table, datasette):
 
 
 @hookspec
-def extra_body_script(template, database, table, datasette):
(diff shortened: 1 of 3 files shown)


russss:
Actually I think this is not the whole story because of the rowid issue. I'm going to think about this one a bit more.

russss:
I think the best approach to this is to pass through the `view_name` parameter I added in #441. It's then simple enough for me to add `.geojson` to the URL in JS - I don't need the pkey.

PR #442, opened 2019-04-29
Suppress rendering of binary data

russss:
Binary columns (including spatialite geographies) get shown as ugly
binary strings in the HTML by default. Nobody wants to see that mess.

Show the size of the column in bytes instead. If you want to decode
the binary data, you can use a plugin to do it.


simonw:
I'm going to merge this and add a unit test.

simonw:
Demo here: https://latest.datasette.io/fixtures/binary_data

I slightly tweaked the copy:

image

PR #447, opened 2019-05-03
Use dist: xenial and python: 3.7 on Travis

simonw:


simonw:
This appears to fix a very weird error we were getting just on Python 3.7-dev: https://travis-ci.org/simonw/datasette/jobs/527858613

That weird error boiled down to `count` being `None`: 

```
{
    "columns": ["pk", "distance", "frequency"],
    "name": "units",
    "count": 3,
    "hidden": False,
    "foreign_keys": {"incoming": [], "outgoing": []},
    "fts_table": None,
    "primary_keys": ["pk"],
}
# compared to:
{
    "name": "units",
    "columns": ["pk", "distance", "frequency"],
    "primary_keys": ["pk"],
    "count": None,
    "hidden": False,
    "fts_table": None,
    "foreign_keys": {"incoming": [], "outgoing": []},
}
```

PR #450, opened 2019-05-04
Coalesce hidden table count to 0

russss:
For some reason I'm hitting a `None` here with a FTS table. I'm not
entirely sure why but this makes the logic work the same as with
non-hidden tables.


russss:
For a bit more context: this fixes a crash with `unsupported operand type(s) for +: 'int' and 'NoneType'` on the index page for me.

simonw:
I fixed this in https://github.com/simonw/datasette/commit/de005b9b7d3db72375e6b8b048d1616a98e6347a instead (hadn't seen this PR)

PR #458, opened 2019-05-09
setup: add tests to package exclusion

### LINKED ISSUE #456: Installing installs the tests package
Because `setup.py` uses `find_packages` and `tests` is on the top-level, `pip install datasette` will install a top-level package called `tests`, which is probably not desired behavior.

The offending line is here:
https://github.com/simonw/datasette/blob/bfa2ae0d16d39bb82dbe4da4f3fdc3c7f6257418/setup.py#L40

And only `pip uninstall datasette` with a conflicting package would warn you by default; apparently another package had the same problem, which is why I get this message when uninstalling:

```
$ pip uninstall datasette
Uninstalling datasette-0.27:
  Would remove:
    /usr/local/bin/datasette
    /usr/local/lib/python3.7/site-packages/datasette-0.27.dist-info/*
    /usr/local/lib/python3.7/site-packages/datasette/*
    /usr/local/lib/python3.7/site-packages/tests/*
  Would not remove (might be manually added):
    [ .. snip .. ]
Proceed (y/n)? 
```

This should be a relatively simple fix, and I could drop a PR if desired!

Cheers

hellerve:
This PR fixes #456 by adding `tests` to the package exclusion list.

Cheers

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     author="Simon Willison",
     license="Apache License, Version 2.0",
     url="https://github.com/simonw/datasette",
-    packages=find_packages(),
+    packages=find_packages(exclude='tests'),
     package_data={"datasette": ["templates/*.html"]},
     include_package_data=True,
     install_requires=[


simonw:
Thanks!

abeyerpath:
*Bumping* this, as the previous fix is passing the wrong type, and not actually addressing the issue...

The `exclude` argument needs an iterable of packages instead of a single string (but since `str` is iterable, it's currently excluding packages `t`, `e`, and `s`.)

PR #480, opened 2019-05-19
Split pypi and docker travis tasks

### LINKED ISSUE #478: Make it so Docker build doesn't delay PyPI release
Datasette automated releases currently include building a Docker image that has a full custom-compiled version of SQLite and SpatiaLite. This takes ages!

I still want to publish this Docker image (to https://hub.docker.com/r/datasetteproject/datasette/tags ) but I'd like it if this wasn't a blocker on pushing the new package to PyPI. Ideally PyPI publish would happen first.

glasnt:
Resolves #478 

This *should* work, but because this is a change that'll only really be testable on a) this repo, b) master branch, this might fail fast if I didn't get the configurations right. 

Looking at #478 it should just be as simple as splitting out the docker and pypi processes into separate jobs, but it might end up being more complicated than that, depending on what pre-processes the pypi deployment needs, and how travisci treats deployment steps without scripts in general.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,17 @@ jobs:
     - stage: release tagged version
       if: tag IS present
       python: 3.6
+      deploy:
+        - provider: pypi
+          user: simonw
+          distributions: bdist_wheel
+          password: ${PYPI_PASSWORD}
+          on:
+            branch: master
+            tags: true
+    - stage: publish docker image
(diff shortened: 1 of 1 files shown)


simonw:
Thanks for this - looks like exactly what I wanted. I'll test this next time I ship a minor release.

PR #518, opened 2019-06-23
Port Datasette from Sanic to ASGI + Uvicorn

simonw:
Most of the code here was fleshed out in comments on #272 (Port Datasette to ASGI) - this pull request will track the final pieces:

- [x] Update test harness to more correctly simulate the `raw_path` issue
- [x] Use `raw_path` so table names containing `/` can work correctly
- [x] Bug: JSON not served with correct content-type
- [x] Get ?_trace=1 working again
- [x] Replacement for `@app.listener("before_server_start")`
- [x] Bug: `/fixtures/table%2Fwith%2Fslashes.csv?_format=json` downloads as CSV
- [x] Replace Sanic request and response objects with my own classes, so I can remove Sanic dependency
- [x] Final code tidy-up before merging to master


datasette/app.py
```diff
@@ -126,8 +127,15 @@
 DEFAULT_CONFIG = {option.name: option.default for option in CONFIG_OPTIONS}
 
 
-async def favicon(request):
-    return response.text("")
+async def favicon(scope, recieve, send):
```
simonw: This should use `asgi_send()` instead.

datasette/app.py
```diff
@@ -588,134 +584,132 @@ class TracingSanic(Sanic):
...
+                info.update(
+                    {"ok": False, "error": message, "status": status, "title": title}
+                )
+                headers = {}
+                if outer_self.cors:
+                    headers["Access-Control-Allow-Origin"] = "*"
+                if scope["path"].split("?")[0].endswith(".json"):
+                    await asgi_send_json(send, info, status=status, headers=headers)
+                else:
+                    template = outer_self.jinja_env.select_template(templates)
```
simonw: The only reason this thing is a nested class is to allow for this access to the `jinja_env` - I could pass this in to the constructor instead.

tests/test_csv.py
```diff
@@ -68,14 +68,14 @@ def test_custom_sql_csv(app_client):
         "/fixtures.csv?sql=select+content+from+simple_primary_key+limit+2"
     )
     assert response.status == 200
-    assert "text/plain; charset=utf-8" == response.headers["Content-Type"]
+    assert "text/plain; charset=utf-8" == response.headers["content-type"]
```
simonw: I'm not entirely happy with these changes regarding header case-insensitivity. It would be good to have a header abstraction which dealt with case-insensitivity in a nicer way.

tests/test_api.py
```diff
@@ -1603,6 +1612,7 @@ def test_infinity_returned_as_invalid_json_if_requested(app_client):
     ] == response.json
 
 
+@pytest.mark.skip
```
simonw: Here's where I'm skipping the test for `?_trace=` until I've fixed it.

simonw:
This is strange: on my local machine `http://127.0.0.1:8001/fixtures/table%2Fwith%2Fslashes.csv` is returning a 404 BUT there's a test for that which is passing under pytest:

https://github.com/simonw/datasette/blob/d60fbfcae2658e71cab6d7b3b9f53f8d895064ef/tests/test_api.py#L721-L727

simonw:
Mystery solved: that's because I'm constructing my own scope object and testing via `ApplicationCommunicator` rather than exercising Uvicorn directly.

https://github.com/simonw/datasette/blob/d60fbfcae2658e71cab6d7b3b9f53f8d895064ef/tests/fixtures.py#L42-L57

I don't want to introduce the complexity of launching a real Uvicorn as part of the tests, so I guess I'll have to carefully update my `ApplicationCommunicator` test harness to more correctly emulate real life.

simonw:
Weird new bug: http://127.0.0.1:8001/fixtures/table%2Fwith%2Fslashes.csv?_format=json is downloading CSV for me now.

https://latest.datasette.io/fixtures/table%2Fwith%2Fslashes.csv?_format=json does the right thing.

simonw:
Another bug: JSON is being served without a content-type header:
```
~ $ curl -i 'http://127.0.0.1:8001/fivethirtyeight/ahca-polls%2Fahca_polls.json'
HTTP/1.1 200 OK
date: Sun, 23 Jun 2019 16:04:01 GMT
server: uvicorn
referrer-policy: no-referrer
transfer-encoding: chunked

{"database": "fivethirtyeight", "table": "ahca-polls/ahca_polls", ...
```

simonw:
OK, for **Get ?_trace=1 working again**. The old code lives in two places:

https://github.com/simonw/datasette/blob/35429f90894321eda7f2db31b9ea7976f31f73ac/datasette/app.py#L546-L560

And then:

https://github.com/simonw/datasette/blob/35429f90894321eda7f2db31b9ea7976f31f73ac/datasette/app.py#L653-L672

So it's stashing something on the request to tell the rest of the code it should be tracing, then using that collected data from the request to add information to the final body.

One possible shape for the replacement is a new ASGI middleware that wraps everything else. We don't have a mutable request object here though, so we will need to untangle this entirely from the request object.

simonw:
** Replacement for @app.listener("before_server_start")** - this is what the [ASGI lifespan protocol](https://asgi.readthedocs.io/en/latest/specs/lifespan.html) is for.

I know Uvicorn supports this because it keeps saying `ASGI 'lifespan' protocol appears unsupported` on the console.

I think the solution here will be to introduce another ASGI wrapper class similar to `AsgiTracer`. I'll model this on the example in the ASGI lifespan spec.

tests/test_api.py
simonw: Fixed this and removed that line.

datasette/app.py
simonw: Fixed.

tests/test_csv.py
simonw: Filed as a separate issue: #522

simonw:
The big one: **Replace Sanic request and response objects with my own classes, so I can remove Sanic dependency**

simonw:
The `InvalidUsage` exception is thrown by Sanic when it gets an unhandled request - usually a HEAD. It was added in efbb4e83374a2c795e436c72fa79f70da72309b8

I'm going to replace it with specific handling for HEAD requests plus a unit test.

simonw:
The Sanic stuff I import at the moment is:

- [ ] `from sanic.exceptions import NotFound`
- [ ] `from sanic.views import HTTPMethodView`
- [ ] `from sanic.request import Request`
- [ ] `from sanic import response` - used for `response.text("")`, `response.redirect(path)`, `response.HTTPResponse` and `response.html()`
- [ ] `from sanic.request import RequestParameters`

datasette/app.py
simonw: Fixed.

simonw:
For the request object.... what are the fields of it I actually use?

* `request.url`
* `request.query_string`
* `request.path`
* `request.method`
* `request.args`
* `request.raw_args`

ALL of those are things that can be derived from the `scope` - so I think my new `Request` class (in `utils/asgi.py`) is just going to be a wrapper around a `scope`.

datasette/utils/asgi.py
```diff
@@ -0,0 +1,344 @@
...
+    @property
+    def args(self):
+        return RequestParameters(parse_qs(qs=self.query_string))
+
+    @property
+    def raw_args(self):
+        return {key: value[0] for key, value in self.args.items()}
+
+    @classmethod
+    def from_path_with_query_string(
```
simonw: I'm sure I could come up with a snappier name for this method. How about one of these?

* `Request.mock("/path?foo=bar")`
* `Request.stub("/path?foo=bar")`
* `Request.fake("/path?foo=bar")`
* `Request.for("/path?foo=bar")`

simonw: https://martinfowler.com/articles/mocksArentStubs.html suggests I should use `fake` here:

Mocks_Aren_t_Stubs

simonw: Renamed to `Request.fake(...)`

simonw:
Last thing is to replace `sanic.response`:

* `response.text("")`
* `response.html()`
* `response.redirect(path)`
* `response.HTTPResponse`

Implementations here: https://github.com/huge-success/sanic/blob/0.7.0/sanic/response.py#L175-L285

simonw:
It's live! https://a559123.datasette.io/

Issue #516, opened 2019-06-22
Enforce import sort order with isort


simonw:
I want to use isort to order imports. A few steps here:

- [ ] Add a .isort.cfg file (see below)
- [ ] Use `isort -rc` to reformat existing code
- [ ] Commit this change
- [ ] Add a unit test that ensures future changes remain isort compatible

simonw:
According to [the black documentation](https://black.readthedocs.io/en/stable/the_black_code_style.html?highlight=isort.cfg) the following config file is necessary to avoid isort and black getting into an edit war with each other:
```
[settings]
multi_line_output=3
include_trailing_comma=True
force_grid_wrap=0
use_parentheses=True
line_length=88
```

simonw:
This test added do `test_black.py` seems to do the trick:
```python
@pytest.mark.parametrize("path", code_root.glob("**/*.py"))
def test_isort(path):
    # Have to capture stdout because isort uses print() directly
    stdout = sys.stdout
    sys.stdout = io.StringIO()
    result = isort.SortImports(path, check=True)
    assert (
        not result.incorrectly_sorted
    ), "{} has incorrectly sorted imports, fix with 'isort -rc . && black .'"
    sys.stdout = stdout
```

simonw:
I'm going to resist the temptation to do this until I've merged the `asgi` branch from #272

simonw:
The Starlette lint and test scripts do this, and also apply autoflake to remove any unnecessary imports: https://github.com/encode/starlette/tree/8c8cc2ec0a5cb834a9a15b871ae8b480503abb67/scripts

simonw:
I just tried this again today by dropping this into `.isort.cfg`:
```ini
[settings]
multi_line_output=3
include_trailing_comma=True
force_grid_wrap=0
use_parentheses=True
line_length=88
known_first_party=datasette
```
And running this in the root of the project:
```
isort .
```
It produced a huge diff, but when I ran the tests:
```bash
pytest -n auto
```
The tests all passed. I'll push a PR.

simonw:
Browse this branch to see the result: https://github.com/simonw/datasette/tree/isort-22-aug-2023

simonw:
Thinking about this again today. Posted about it on Discord: https://discord.com/channels/823971286308356157/823971286941302908/1143729300349132933

I won't enforce it in a `pytest` test, I'll enforce it with a CI lint check that's also in the `Justfile` here instead:

[...]

PR #524, opened 2019-06-24
Sort commits using isort, refs #516

simonw:
Also added a lint unit test to ensure they stay sorted. #516


simonw:
This branch is WAY out of date.

Issue #541, opened 2019-07-05
Plugin hook for adding extra template context variables


simonw:
It turns out I need this for https://github.com/simonw/datasette-auth-github/issues/5

It can be modelled on the `extra_body_script` hook: https://datasette.readthedocs.io/en/stable/plugins.html#extra-body-script-template-database-table-view-name-datasette

simonw:
Example usage:
```python
@hookimpl
def extra_template_vars(request):
    return {
        "auth": request.scope.get("auth"),
    }
```

simonw:
It should allow for awaitables too, to allow for execution of SQL. For example:

```python
@hookimpl
def extra_template_vars(datasette):
    async def inner():
        first_db = list(datasette.databases.keys())[0]
        return {
            "sqlite_version": (
                await datasette.execute(first_db, "select sqlite_version()")
            ).rows[0][0]
        }
    return inner
```

PR #542, opened 2019-07-05
extra_template_vars plugin hook

simonw:
Refs #541

### DIFF
--- a/datasette/hookspecs.py
+++ b/datasette/hookspecs.py
@@ -36,6 +36,11 @@ def extra_body_script(template, database, table, view_name, datasette):
 
 
 @hookspec
+def extra_template_vars(template, database, table, view_name, request, datasette):
+    "Extra template variables to be made available to the template - can return dict or callable or awaitable"
+
+
+@hookspec
 def publish_subcommand(publish):
     "Subcommands for 'datasette publish'"
 
--- a/datasette/views/base.py
+++ b/datasette/views/base.py
@@ -102,7 +102,7 @@ class BaseView(AsgiView):
     def database_color(self, database):
         return "ff0000"
 
-    def render(self, templates, **context):
+    async def render(self, templates, request, context):
         template = self.ds.jinja_env.select_template(templates)
         select_templates = [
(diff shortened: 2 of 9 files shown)


simonw:
I'm getting this weird failure again:
```
self = <click.testing.CliRunner object at 0x7f4295b7bd30>
    @contextlib.contextmanager
    def isolated_filesystem(self):
        """A context manager that creates a temporary folder and changes
        the current working directory to it for isolated filesystem tests.
        """
>       cwd = os.getcwd()
E       FileNotFoundError: [Errno 2] No such file or directory
```
I saw this the other day too, but managed to avoid it by skipping any code that messed around with temporary files.

Baffled as to why this is showing up now, as none of the changes I'm making here have any relevance to the way temporary files are created or the current working directory is set.

I understand the root cause here to be that the working directory for the tests was deleted while the tests were still running, which would seem to indicate that something to do with a `tmpdir` fixture is at fault.

simonw:
I bet that error is something to do with this:
https://github.com/simonw/datasette/blob/e06e083240522a281b280b6461e935bf4b47bdbc/tests/test_plugins.py#L195-L197

simonw:
I've confirmed that it's the `template_dir=` line quoted above. It isn't the fault of `pathlib` - if you hard-code in the path string like this I still get the same failures:
```
    template_dir = "/Users/simonw/Dropbox/Development/datasette/tests/test_templates"
    for client in make_app_client(
        template_dir=template_dir
    ):
```
So for some reason setting `template_dir` to a concrete directory breaks those other tests.

simonw:
Here's what `isolated_filesystem()` does:
>     A context manager that creates a temporary folder and changes the current working directory to it for isolated filesystem tests.
https://github.com/pallets/click/issues/395 is a four year old issue suggesting allowing it to have a parent tmp directory passed to it, but since that issue was never closed I don't think it's going to help me.

simonw:
This looks relevant: https://github.com/pytest-dev/pytest/issues/1235 - "pytest explodes if getcwd fails"

Issue #510, opened 2019-06-15
Ability to facet by delimiter (e.g. comma separated fields)


simonw:
E.g. if a field contains "Tags,With,Commas" be able to facet them in the same way as `_facet_arry=` lets you facet `["Tags", "With", "Commas"]`

simonw:
Part of #551

PR #546, opened 2019-07-07
Facet by delimiter

simonw:
Refs #510


simonw:
Still needs tests and documentation.

simonw:
Also maybe this should be added as a suggested facet if a column clearly contains commas that, when split upon, would result in a array recommendation?

PR #554, opened 2019-07-09
Fix static mounts using relative paths and prevent traversal exploits

abdusco:
While debugging why my static mounts using a relative path (`--static mystatic:rel/path/to/dir`) not working, I noticed that the requests fail no matter what, returning 404 errors. 

The reason is that datasette tries to prevent traversal exploits by checking if the path is relative to its registered directory. This check fails when the mount is a relative directory, because `/abs/dir/file` obviously not under `dir/file`. 

https://github.com/simonw/datasette/blob/81fa8b6cdc5457b42a224779e5291952314e8d20/datasette/utils/asgi.py#L303-L306

This also has the consequence of returning any requested file, because when `/abs/dir/../../evil.file` resolves `aiofiles` happily returns it to the client after it resolves the path itself. The solution is to make sure we're checking relativity of paths after they're fully resolved.

I've implemented the mentioned changes and also updated the tests.

### DIFF
--- a/datasette/utils/__init__.py
+++ b/datasette/utils/__init__.py
@@ -735,7 +735,7 @@ class StaticMount(click.ParamType):
                 param,
                 ctx,
             )
-        path, dirpath = value.split(":")
+        path, dirpath = value.split(":", 1)
         dirpath = os.path.abspath(dirpath)
         if not os.path.exists(dirpath) or not os.path.isdir(dirpath):
             self.fail("%s is not a valid directory path" % value, param, ctx)
--- a/datasette/utils/__init__.py
+++ b/datasette/utils/__init__.py
@@ -736,6 +736,7 @@ class StaticMount(click.ParamType):
                 ctx,
             )
         path, dirpath = value.split(":")
+        dirpath = os.path.abspath(dirpath)
         if not os.path.exists(dirpath) or not os.path.isdir(dirpath):
             self.fail("%s is not a valid directory path" % value, param, ctx)
         return path, dirpath
--- a/datasette/utils/asgi.py
+++ b/datasette/utils/asgi.py
@@ -300,7 +300,7 @@ async def asgi_send_file(
 def asgi_static(root_path, chunk_size=4096, headers=None, content_type=None):
     async def inner_static(scope, receive, send):
         path = scope["url_route"]["kwargs"]["path"]
-        full_path = (Path(root_path) / path).absolute()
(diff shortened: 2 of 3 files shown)


abdusco:
I've also added another fix for using static mounts with absolute paths on Windows.

abdusco:
I wanted to add a test for it too, but I've realized it's impossible to test a server process as we cannot get its exit code.

```python
# tests/test_cli.py
def test_static_mounts_on_windows():
    if sys.platform != "win32":
        return
    runner = CliRunner()
    result = runner.invoke(
        cli, ["serve", "--static", r"s:C:\\"]
    )
    assert result.exit_code == 0
```

simonw:
Thanks for this!

The tests are failing for Python 3.5 right now which is strange. https://travis-ci.org/simonw/datasette/jobs/556272017

One failure looks like this:
```
app_client = <tests.fixtures.TestClient object at 0x7fb7803285f8>
    def test_static(app_client):
        response = app_client.get("/-/static/app2.css")
>       assert response.status == 404
E       assert 500 == 404
E        +  where 500 = <tests.fixtures.TestResponse object at 0x7fb780328a58>.status
/home/travis/build/simonw/datasette/tests/test_html.py:56: AssertionError
----------------------------- Captured stderr call -----------------------------
Traceback (most recent call last):
  File "/home/travis/build/simonw/datasette/datasette/utils/asgi.py", line 100, in __call__
    return await view(new_scope, receive, send)
  File "/home/travis/build/simonw/datasette/datasette/utils/asgi.py", line 303, in inner_static
    full_path = (Path(root_path) / path).resolve().absolute()
  File "/opt/python/3.5.6/lib/python3.5/pathlib.py", line 1109, in resolve
    s = self._flavour.resolve(self)
  File "/opt/python/3.5.6/lib/python3.5/pathlib.py", line 330, in resolve
    return _resolve(base, str(path)) or sep
  File "/opt/python/3.5.6/lib/python3.5/pathlib.py", line 315, in _resolve
    target = accessor.readlink(newpath)
  File "/opt/python/3.5.6/lib/python3.5/pathlib.py", line 422, in readlink
    return os.readlink(path)
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/simonw/datasette/datasette/static/app2.css'
```
Maybe an exception was renamed between 3.5 and 3.6?

simonw:
Released as 0.29.1: https://datasette.readthedocs.io/en/latest/changelog.html#v0-29-1

PR #556, opened 2019-07-09
Add support for running datasette as a module

abdusco:
This PR allows running datasette using `python -m datasette` command in addition to just running the executable.

This function is quite useful when debugging a plugin in a project because IDEs like PyCharm can easily start a debug session when datasette is run as a module in contrast to trying to attach a debugger to a running process.

image


simonw:
This is a really neat trick, thanks!

PR #578, opened 2019-09-29
Added support for multi arch builds

heussd:
Minor changes in Dockerfile and new Makefile to support Docker multi architecture builds. `make`will build one image per architecture and push them as one Docker manifest to Docker Hub. Feel free to change `IMAGE_NAME ` to `datasetteproject/datasette` to update your official Docker Hub image(s).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ COPY --from=build /usr/local/lib/ /usr/local/lib/
 # Copy executables
 COPY --from=build /usr/local/bin /usr/local/bin
 # Copy spatial extensions
-COPY --from=build /usr/lib/x86_64-linux-gnu /usr/lib/x86_64-linux-gnu
+COPY --from=build /usr/lib/*-linux-gnu /usr/lib/
 
 ENV LD_LIBRARY_PATH=/usr/local/lib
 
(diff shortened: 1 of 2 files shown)


simonw:
So this is a new Docker feature from April 2019? https://engineering.docker.com/2019/04/multi-arch-images/

> So why buildx?  Let’s first start with the name.  The x stands for experimental.  In the future, when these new build features are stable and made generally available, we will drop the x and integrate these features directly into the existing docker build command.  Note that as buildx is experimental, features and flags are subject to change.

Is there a strong use-case for supporting this while it's still experimental? Users who want Datasette to run on ARM can build their own image pretty easily, so I'm not sure if it's worth depending on an experimental feature for the core Datasette project.

Is there any indication from the Docker team as to when this experimental feature might stop being marked as experimental?

heussd:
My use case was: I wanted to use datasette on a Raspberry Pi. `docker pull datasetteproject/datasette` pulled the official image, which then failed to execute because it is not ARM ready. Building my own quite took some time (~60 minutes via Qemu on Intel i5). 

You are right, the build method is quite new and I would not be surprised if the syntax / command will change in future. The outcome however, a Docker multi-architecture manifest, is aligned with Docker's strategy on how to tackle multiple architectures: transparently, on the registry-side.

I just thought it would be nice to have the official image ready for multiple architectures. But I fully understand if the current methods feel too experimental to be mergable...

simonw:
Yes I'm going to hold of on this until the Docker feature is less experimental.

PR #579, opened 2019-10-02
New connection pooling plus .execute() refactor

simonw:


simonw:
I'm going to extract out the refactoring of `.execute()` from `Datasette` to `Database` into a separate commit.

PR #590, opened 2019-10-11
Handle spaces in DB names

### LINKED ISSUE #503: Handle SQLite databases with spaces in their names?
I named my SQLite database "Government workers" and published it to Heroku. When I clicked the "Government workers" database online it lead to a 404 page: `Database not found: Government%20workers`.

I believe this is because the database name has a space.

rixx:
Closes #503

### DIFF
--- a/datasette/views/base.py
+++ b/datasette/views/base.py
@@ -203,12 +203,13 @@ class DataView(BaseView):
                 hash = None
         else:
             name = db_name
-        # Verify the hash
+        name = urllib.parse.unquote_plus(name)
         try:
             db = self.ds.databases[name]
         except KeyError:
             raise NotFound("Database not found: {}".format(name))
 
+        # Verify the hash
         expected = "000"
         if db.hash is not None:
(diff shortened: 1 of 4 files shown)


simonw:
We need some unit tests for this.

I'd be fine with us renaming `extra_database.db` to `extra database.py` in these existing fixtures:

https://github.com/simonw/datasette/blob/af2e6a5cf186a7200d76cb67ac30fa59cc24d84e/tests/fixtures.py#L174-L185

Then we could add an extra `test_database_name_with_spaces(...)` test in `test_api.py` which uses parametrize to exercise a bunch of URLs and check they all return a 200 - something a bit like this but covering a range of URL paths (one for database, table, row and table with filters):

https://github.com/simonw/datasette/blob/90d4f497f9b3f6a5882937c91fddb496ac3e7368/tests/test_api.py#L746-L765

rixx:
Ah, thank you – I saw the need for unit tests but wasn't sure what the best way to add one would be.

rixx:
Added tests.

PR #591, opened 2019-10-11
Sort databases on homepage by argument order

### LINKED ISSUE #585: Databases on index page should display in order they were passed to "datasette serve"?
If you run this:

    datasette serve -h 127.0.0.1 -p 8000 -m phone-locations.db healthkit.db locations.db genome.db

Then the index page for that Datasette instance should show the databases in the order they were specified on the command-line.

Mind you when we add pagination to that page in #468 we may want to do something different here.

rixx:
Closes #585


simonw:
Neat solution.

PR #595, opened 2019-10-13
bump uvicorn to 0.9.0 to be Python-3.8 friendly

stonebig:
as uvicorn-0.9 is needed to get websockets-8.0.2, which is needed to have Python-3.8 compatibility

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         "hupper~=1.0",
         "pint~=0.8.1",
         "pluggy~=0.12.0",
-        "uvicorn~=0.8.4",
+        "uvicorn~=0.9.0",
         "aiofiles~=0.4.0",
     ],
     entry_points="""


simonw:
Ooh this is tricky... https://travis-ci.org/simonw/datasette/jobs/597216285 failed because it looks like `websockets-8.0.2` isn't compatible with Python 3.5

This is a problem. I need to keep Datasette working with 3.5 for compatibility with Glitch. But obviously I also want it to work with 3.8. 

I wonder if there's a smart solution here. @tomchristie any thoughts?

stonebig:
Maybe make the setup rule conditional, so that below python-3.6, it looks for unicorn-8 ?

tomchristie:
🤷‍♂️ @stonebig's suggestion would be the best I got too, *if* you want to support 3.5->3.8.

It's either that, or hold off on 3.8 support until you're ready to go to 3.6->3.8.

simonw:
There more discussion around this on the Glitch forums here: https://support.glitch.com/t/can-you-upgrade-python-to-latest-version/7980

simonw:
I like the conditional dependency for the moment - maybe until 3.5 becomes officially unsupported.

simonw:
Glitch has been upgraded to Python 3.7. I think I'm happy to drop 3.5 support now - users who want Python 3.5 can get it by installing `datasette==0.30.2`

simonw:
#622

tomchristie:
> Glitch has been upgraded to Python 3.7.

Whoop! 🥳 ✨

simonw:
Closing this in favour of #627 which upgrades to uvicorn 0.10.4

Issue #600, opened 2019-10-18
Don't auto-format SQL on first page load


simonw:
I've gone back and forth on this a bit, but I've decided I'm not keen on the way Datasette now automatically formats SQL when a query (or canned query) page first loads.

I like having an optional "Format SQL" button, but applying formatting automatically means that if the user has carefully formatted their SQL to a specific style their formatting will be automatically over-ridden.

PR #601, opened 2019-10-18
Don't auto-format SQL on page load

simonw:
Refs #600


simonw:
@rixx does this look like the right way to do #600 to you?

rixx:
That looks right, and I completely agree with the intent.

rixx:
The only negative impact that comes to mind is that now you have no way to get the read-only query to be formatted nicely, I think, so maybe a second PR adding the formatting functionality even to the read-only page would be good?

simonw:
That's a good idea. It could even be smart about it and only show the "Format SQL" button on that page of the output of the function would differ from what's there already.

rixx:
Submitted in #602!

PR #602, opened 2019-10-20
Offer to format readonly SQL

rixx:
Following discussion in #601, this PR adds a "Format SQL" button to
read-only SQL (if the SQL actually differs from the formatting result).

It also removes a console error on readonly SQL queries.

### DIFF
--- a/datasette/templates/_codemirror_foot.html
+++ b/datasette/templates/_codemirror_foot.html
@@ -6,21 +6,32 @@ window.onload = () => {
     if (sqlFormat && !readOnly) {
         sqlFormat.hidden = false;
     }
-    var editor = CodeMirror.fromTextArea(sqlInput, {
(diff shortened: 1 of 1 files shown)


simonw:
Live demo can now be seen here: https://latest.datasette.io/fixtures/%F0%9D%90%9C%F0%9D%90%A2%F0%9D%90%AD%F0%9D%90%A2%F0%9D%90%9E%F0%9D%90%AC

simonw:
I wonder if the `Format SQL` button should disappear in the read-only case once it has been clicked? It won't do anything if it's clicked again and there's no textarea for the user to mess up the now-formatted SQL.

rixx:
Not sure – I'm always a bit weirded out when elements that I clicked disappear on me.

PR #644, opened 2019-11-30
Validate metadata json on startup

### LINKED ISSUE #260: Validate metadata.json on startup
It's easy to misspell the name of a database or table and then be puzzled when the metadata settings silently fail.

To avoid this, let's sanity check the provided metadata.json on startup and quit with a useful error message if we find any obvious mistakes.

chris48s:
This PR adds a sanity check which builds up a marshmallow schema on-the-fly based on the structure of the database(s) on startup and then validates the metadata json against it.

In case of invalid data, this will raise with a descriptive error e.g:

```
marshmallow.exceptions.ValidationError: {'databases': {'fixtures': {'tables': {'not_a_table': ['Unknown field.']}}}}
```

Closes #260

---

This was intended to be fairly self-contained, but then while I was working on it, I hit some problems getting the tests to pass in the context of the test suite as a whole. My tests passed in isolation, but then failed while doing a full test suite run. That's when the worms started coming out of the can :bug: After some sleuthing, it turned out this was essentially the result of several issues intersecting:

* There are certain events in the application lifecycle where the metadata schema can be modified after it is loaded e.g: https://github.com/simonw/datasette/blob/a562f2965552fb2dbbbd74df245c9965ee23d886/datasette/app.py#L299-L320 This means that sometimes what goes in isn't always exactly what comes out when you call `/-/metadata`.
* Because the test fixtures use session scope for performance reasons if one unit test performs an action which mutates the metadata, that can impact on other unit tests which run after it using the same fixture.
* Because the `self._metadata` property was being set with a simple assignment `self._metadata = metadata`, that created an object reference to the test fixture data, so operating on `self._metadata` was actually modifying the test fixture `METADATA` meaning that depending on when it was loaded in the test suite lifecycle, `METADATA` had different content, which was somewhat unexpected.

As such, I've added some band-aids in 3552024 and 6859fd8:
* Switching the metadata object to a `deepcopy` of the input prevents us directly mutating the input fixture.
* I've switched some of the tests to use a fixture with function scope instead of session scope so we're working on a clean copy that hasn't been mutated by other tests where necessary but keeping session scope in most cases for performance.
* I haven't really addressed the fact that sometimes the metadata object gets mutated in place, so the object that is served from `/-/metadata` isn't necessarily always exactly the same as the file you fed into it on init. I'm not sure how much of a problem that is. The way the tests were written makes me think it was unexpected, but getting into it feels like too much scope creep for this PR so its probably best addressed as another issue.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -6,6 +6,7 @@ import sys
 import threading
 import traceback
 import urllib.parse
+from copy import deepcopy
 from concurrent import futures
 from pathlib import Path
 
@@ -173,7 +174,7 @@ class Datasette:
             self.databases[db.name] = db
         self.cache_headers = cache_headers
         self.cors = cors
-        self._metadata = metadata or {}
+        self._metadata = deepcopy(metadata) or {}
         self.sqlite_functions = []
         self.sqlite_extensions = sqlite_extensions or []
         self.template_dir = template_dir
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -21,6 +21,7 @@ from .views.special import JsonDataView
 from .views.table import RowView, TableView
 from .renderer import json_renderer
 from .database import Database
+from .metadata import get_metadata_schema
 
 from .utils import (
     QueryInterrupted,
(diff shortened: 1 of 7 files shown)


chris48s:
Hi @simonw

Have you had a chance to look at this at all?

I'm going to have a chunk of time free next week so if there is additional work needed on this, that would be a particularly convenient time for me to revisit this.

Cheers

PR #652, opened 2019-12-18
Quick (and uninformed and perhaps misguided) attempt to add a <base> url for hosting datasette at a particular host/URI

### LINKED ISSUE #394: base_url configuration setting
I've identified a couple of use-cases for running Datasette in a way that over-rides the default way that internal URLs are generated.

1. Running behind a reverse proxy. I tried running Datasette behind a proxy and found that some of the generated internal links incorrectly referenced `http://127.0.0.1:8001/fixtures/...` - when they should have been referencing `http://my-host.my-domain.com/fixtures/...` - this is a problem both for links within the HTML interface but also for the `toggle_url` keys returned in the JSON as part of the facets datastructure.
2. I would like it to be possible to host a Datasette instance at e.g. `https://www.mynewspaper.com/interactives/2018/election-results/` - either through careful HTTP proxying or, once Datasette has been ported to ASGI, by mounting a Datasette ASGI instance deep within an existing set of URL routes.

I'm going to add a `url_prefix` configuration option. This will default to `""`, which means Datasette will behave as it does at the moment - it will use `/` for most URL prefixes in the HTML version, and an absolute URL derived from the incoming `Host` header for URLs that are returned as part of the JSON output.

If `url_prefix` is set to another value (either a full URL or a path) then this path will be appended to all generated URLs.

terrycojones:
As usual, I don't really know what I'm doing...  so this is just a suggested approach. I've not written tests, I've not run the tests, I don't know if I've missed some absolute URLs that would need to have the leading slash dropped.

BUT, I tested it with `--config base_url:http://127.0.0.1:8001/` on the command line and from what little I know about datasette it's at least working in some obvious cases.

My changes are based on what I saw in https://github.com/simonw/datasette/commit/8da2db4b71096b19e7a9ef1929369b8483d448bf (thanks!)

I'm happy to be more thorough on this if you think it's worth pursuing.

Fixes #394  (he said, optimistically).


datasette/app.py
```diff
@@ -126,6 +126,11 @@
         False,
         "Force URLs in API output to always use https:// protocol",
     ),
+    ConfigOption(
+        "base_url",
+        "",
```
terrycojones: Maybe the default should be `/` to try to ensure that the behaviour with this change is in accordance with the way things were. Then the leading slashes that I dropped get put back in by the `<base>` tag in case you don't specify one in the config. The downside(?) is that your pages always have a base tag.  I used to use `<base>` a lot back in 1995 or so, to host multiple domains/sites on one server, before the advent of the `Host` header, etc.

simonw:
This has been replaced by #708, which is close to being ready.

PR #660, opened 2020-01-21
gcloud run is now GA, s/beta//

glasnt:


simonw:
Most excellent fix, thank you.

Issue #577, opened 2019-09-23
Utility mechanism for plugins to render templates


simonw:
Sometimes a plugin will need to render a template for some custom UI. We need a documented API for doing this, which ensures that everything will work correctly if you extend base.html etc.

See also #576. This could be a `.render()` method on the Datasette class, but that feels a bit weird - should that class also take responsibility for rendering?

simonw:
Here's the horrific (stack inspecting) workaround I came up with for a prototype of `datasette-atom`: https://github.com/simonw/datasette-atom/commit/c0e3bd9556d7b31f253a8bf666d42205cd24f4fc#diff-acba9942430bc5e616410567296f92ffR87

simonw:
I think a `.render_template()` method on the `Datasette` class would be acceptable, since the purpose of that class will mainly be to provide a documented API for plugins to use: #576

simonw:
Better idea: take advantage of plushy dependency injection. If a plugin takes a `render` argument we can send it a function that can be used to render a template.

The need to `await render(...)` might be difficult here though.

simonw:
I'm going to go with `Datasette(...).render_template(...)` - I need that for #648, and it makes sense to me that the `Datasette` class is the documented interface that plugins interact with.

simonw:
The reason I've been dragging my heels on adding template rendering to the Datasette class is that it feels messy - should that class be responsible for both data access AND template rendering?

I think I can come to terms with this thanks to plugins. The Datasette class can represent the family of features that plugins affect - which means that having it expose the template rendering API is reasonable.

simonw:
The code in question currently lives in `BaseView.render()`:

https://github.com/simonw/datasette/blob/d54318fc7f2565e6121920ce1ea9cb8b700e629a/datasette/views/base.py#L106-L163

Should `datasette.render_template()` do exactly this, or should it be slightly different?

Plugins need the option to not pass a `request` object - so maybe that parameter becomes optional.

Perhaps plugins should be able to render templates without other plugins getting to inject their own variables?

Does it always make sense to dump in all of those extra template context variables?

simonw:
For the moment I'm going to move it to `async def render_template()` on `datasette` but otherwise keep the implementation the same.

The new signature will be:

    async def render_template(self, template, context=None, request=None, view_name=None):

`template` can be a list of strings or a single string. If a list of strings a template will be selected from them.

I'll reconsider the large list of default context variables later on in a separate ticket.

Issue #576, opened 2019-09-23
Documented internals API for use in plugins


simonw:
Quite a few of the plugin hooks make a `datasette”`instance of the Datasette class available to the plugins, so that they can look up configuration settings and execute database queries.

This means it should provide a documented, stable API so that plugin authors can rely on it.

simonw:
I wrote about this a bit here: https://simonwillison.net/2020/Jan/6/sitemap-xml/

simonw:
OK, I've made a start on this now in 3ffb8f3b98252531d11897fd431711e9b8045ace - still plenty more methods to document. More importantly that class has a LOT of junk methods on that no-one should ever call from a plugin, so I need to decide what to do about those.

simonw:
I need to document the `.databases` property - both how to get access to specific databases and how to access the first/only database.

simonw:
Need to document `datasette.metadata()` - see #780

simonw:
There's a bunch of methods on that class which I could add a `_` prefix to and leave undocumented. `datasette.register_renderers()` should be `datasette._register_renderers()` for example.

simonw:
Here's the current `Datasette` class as introspected using the webserver run by `pydoc -p 8000`:

<img width="956" alt="Pydoc__module_datasette_app" src="https://user-images.githubusercontent.com/9599/83329718-f0835680-a23f-11ea-8cbf-1a1e5dafb7ec.png">

simonw:
**__init__**(self, files, immutables=None, cache_headers=True, cors=False, inspect_data=None, metadata=None, sqlite_extensions=None, template_dir=None, plugins_dir=None, static_mounts=None, memory=False, config=None, version_note=None, config_dir=None)

`Initialize self.  See help(type(self)) for accurate signature.`

**absolute_url**(self, request, path)

**add_database**(self, name, db)

**app**(self)

`Returns an ASGI app function that serves the whole of [Datasette](http://localhost:8066/datasette.app.html#Datasette)`

**app_css_hash**(self)

**config**(self, key)

**config_dict**(self)

**connected_databases**(self)

**execute**(self, db_name, sql, params=None, truncate=False, custom_time_limit=None, page_size=None, log_sql_errors=True)

**expand_foreign_keys**(self, database, table, column, values)

`Returns dict mapping (column, value) -> label`

**get_canned_queries**(self, database_name)

**get_canned_query**(self, database_name, query_name)

[...]

PR #664, opened 2020-02-04
Datasette.render_template() method

simonw:
Refs #577

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1,6 +1,7 @@
 import asyncio
 import collections
 import hashlib
+import itertools
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


simonw:
Needs documentation.

simonw:
More importantly, I need to prototype up a plugin that actually uses this method - probably `datasette-upload-csvs` - to make sure it serves its purpose.

simonw:
It doesn't make sense for extra body scripts to be available for all template renders (including from plugins) but extra CSS and JS to only be handled by the BaseView render method - especially since extra body scripts are likely to depend on extra JS files.

simonw:
I'm going to ship this without documentation, get a couple of plugins out that use it and then add documentation once I'm confident it's right.

simonw:
See #576 for existing ticket covering Datasette class documentation.

PR #666, opened 2020-02-08
Use inspect-file, if possible, for total row count

kevindkeogh:
For large tables, counting the number of rows in the table can take a
signficant amount of time. Instead, where an inspect-file is provided
for an immutable database, look up the row-count for a plain count(*).


simonw:
This is a smart optimization. I'd like to add a test for it, since the logic is a tiny bit more complex.

kevindkeogh:
It was very helpful for me, using it for a 15M row table. Added a test, happy to amend though!

simonw:
This is great, thanks!

PR #669, opened 2020-02-11
fix db-to-sqlite command in ecosystem doc page

adipasquale:
the `--connection` parameter has become positional


simonw:
Thanks for spotting this!

Issue #698, opened 2020-03-16
Ability for a canned query to write to the database


simonw:
Canned queries are currently read-only: https://datasette.readthedocs.io/en/0.38/sql_queries.html#canned-queries

Add a `"write": true` option to their definition in `metadata.json` which turns them into queries that are submitted via POST and send their queries to the write queue.

Then they can be used as a really quick way to define a writable interface and JSON API!

simonw:
```json
{
    "databases": {
       "my-database": {
           "queries": {
               "add_twitter_handle": {
                   "sql": "insert into twitter_handles (username) values (:username)"
               }
           }
       }
    }
}
```

simonw:
By default this will extract the `:params` using the existing regular expression - which can occasionally break if there is a rogue `:` in the rest of the query.

To address this: allow an extra optional `"params": ["usernname"]` field in metadata which, if available, is used instead of the regular expression extraction.

simonw:
This is going to need CSRF protection.

simonw:
Also relevant: this will benefit from an authentication/permissions layer: #699

simonw:
Urgh, this is also going to need me to handle POST form submissions which means I need to be able to parse the form body. I guess that will go in https://github.com/simonw/datasette/blob/master/datasette/utils/asgi.py

simonw:
I have code for form parsing in `asgi-csrf`: https://github.com/simonw/asgi-csrf/blob/3fc164a9258ebfa616d4e724fcb102b117277c6e/asgi_csrf.py#L91-L109

simonw:
It would be useful if this feature supported different types of input - at the most basic level `<input type="text">` v.s. `<textarea>`, but numbers, `<select>` etc would be useful too. This can be added later as additional `metadata.json` settings.

simonw:
This is going to require a pretty full rewrite of the existing canned query logic, which currently shares a codepath with the code that lets you submit a SQL query:

https://github.com/simonw/datasette/blob/298a899e792ebd0cd82a5f01b613c31f19082e51/datasette/views/base.py#L464-L563

simonw:
This feature should include the ability to set a custom redirect URL for after the query has been executed - that way it can be used to build things like "delete this row" which redirect back to the correct table.

simonw:
YAML for metadata is relevant to this: #713

simonw:
I really want the option to use a `<textarea>` for a specific value.

Idea: metadata syntax like this:

```json
[...]

PR #703, opened 2020-03-21
WIP implementation of writable canned queries

simonw:
Refs #698.

Still todo:

- [ ] Unit tests
- [ ] Figure out `.json` mode
- [x] Flash message solution
- [ ] CSRF protection
- [ ] Better error message display on errors
- [ ] Documentation
- [ ] Maybe widgets?


simonw:
I'm going to split canned queries out from arbitrary queries to make the code easier to follow.

simonw:
Adding some still-to-do items to a checklist in the description.

datasette/views/database.py
```diff
@@ -129,12 +131,46 @@ async def data(
...
         templates = ["query-{}.html".format(to_css_class(database)), "query.html"]
+
+        # Execute query - as write or as read
+        if write:
+            if request.method == "POST":
+                params = await request.post_vars()
+                write_ok = await self.ds.databases[database].execute_write(
+                    sql, params, block=True
+                )
+                return self.redirect(request, request.path + '?_success=Query+executed_successfully')
```
simonw: Passing `?_success` like this isn't necessarily the best approach. Potential improvements include:
- Signing this message so it can't be tampered with (I could generate a signing secret on startup)
- Using a cookie with a temporary flash message in it instead
- Using HTML5 history API to remove the `?_success=` from the URL bar when the user lands on the page
If I add an option to redirect the user to another page after success I may need a mechanism to show a flash message on that page as well, in which case I'll need a general flash message solution that works for any page.

simonw: I could add itsdangerous as a signing dependency, which delightfully has zero dependencies of its own: https://pypi.org/project/itsdangerous

simonw:
This pull request got too messy. I'm going to abandon this and start a new one.

PR #708, opened 2020-03-24
base_url configuration setting, refs #394

simonw:
Pull request implementing #394

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -135,7 +135,9 @@ CONFIG_OPTIONS = (
         False,
         "Allow display of template debug information with ?_context=1",
     ),
+    ConfigOption("base_url", "/", "Datasette URLs should use this base"),
 )
+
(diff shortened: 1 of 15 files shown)


simonw:
Spotted a bug: `<link rel="stylesheet" href="/-/static/app.css?ffa51a">` in the templates. I need to ensure the static assets are served from the new location as well (and any plugin assets too).

simonw:
This is now ready for people to try out. Quoting https://github.com/simonw/datasette/issues/394#issuecomment-603525339

> Anyone willing to give it a quick test and see if it works for your particular use-case? You can install it with:
> 
> ```
> pip install https://github.com/simonw/datasette/archive/base-url.zip
> ```
> 
> Then you can run Datasette like this:
> 
> ```
> datasette fixtures.db --config base_url:/new-base/path/here/
> ```

Issue #713, opened 2020-04-02
Support YAML in metadata - metadata.yaml


simonw:
I was originally going to do this with a plugin - see #357 - but the more I work with `metadata.json` the more I want it to just accept YAML as an optional alternative to JSON.

The best example why is still this one: https://github.com/simonw/russian-ira-facebook-ads-datasette/blob/master/russian-ads-metadata.yaml

YAML is just SO much better than JSON for multi-line strings - in particular HTML and SQL, both of which are common in `metadata.json` files.

simonw:
I can add https://pypi.org/project/PyYAML/ as a dependency for this.

One open question: how best to tell the difference between a JSON and a YAML file? I'd rather not do it based on a file extension.

Since I'm going to read the whole file into memory anyway (rather than try to stream it) and I only load it once at startup, maybe I try to parse with one and, if there's a parsing error, try the other one before giving up?

simonw:
I'm going to depend on ~=5.3 https://github.com/yaml/pyyaml/blob/master/CHANGES

simonw:
I'm going to try to parse as JSON first, then fall back to YAML, then error.

simonw:
Don't forget to fix the code in publish that loads and modifies metadata, e.g.:

https://github.com/simonw/datasette/blob/2aaad72789c427875426673c1a43e67c86fc970e/datasette/utils/__init__.py#L361-L367

simonw:
https://github.com/simonw/datasette/blob/6717c719dd36dc2adc0f9da38a8c8e08129e96b4/datasette/utils/__init__.py#L795-L803

simonw:
Documentation: https://datasette.readthedocs.io/en/latest/metadata.html#using-yaml-for-metadata

PR #714, opened 2020-04-02
--metadata accepts YAML as well as JSON

simonw:
Refs #713. Still needs tests and documentation.


simonw:
Writing unit tests for this is tricky as the YAML loading happens in the CLI code, but most of my existing unit tests skip that step and run directly against a Datasette instance.

Issue #648, opened 2019-12-08
Mechanism for adding arbitrary pages like /about


simonw:
For www.niche-museums.com I solved this by creating an empty `about.db` database file - see https://simonwillison.net/2019/Nov/25/niche-museums/

I want a neater mechanism for this.

simonw:
Idea: do this with a simple template naming convention.

If you hit `/about` and there is no matching database, check for a template file called `about-page.html`. If it exists, render it. Otherwise return a 404 database not found.

simonw:
Alternative idea: a new concept of "pages" which live inside `templates/pages/` and where the file name minus the `.html` extension defines the URL.

`templates/about/me.html` would be served at `/about/me` - but only if no matching database and table were found.

This only takes effect on 404 errors from core Datasette.

simonw:
Stretch goal: it would be neat if these pages could return custom HTTP headers (eg content-type) and maybe even status codes (eg for redirects) somehow.

simonw:
The implementation in https://github.com/simonw/datasette/commit/c5e8cd84d3ef55ed86771ac0bde0ca91d6b0e07a acts as a proof of concept. It has a big flaw though: it doesn't reuse the regular render() mechanism, which means it doesn't register custom template tags from plugins.

This is bad because it means that pages rendered in this way cannot take advantage of things like [datasette-template-sql](https://github.com/simonw/datasette-template-sql).

This means this issue is likely dependent on #577 - a documented mechanism to allow plugins to render templates.

simonw:
Now that I've closed #577 this should be pretty easy to implement.

simonw:
> Stretch goal: it would be neat if these pages could return custom HTTP headers (eg content-type) and maybe even status codes (eg for redirects) somehow.

I think I could do that with a custom template function - if that function is called during the render then we follow those instructions instead of returning the rendered HTML.

simonw:
The trickiest part here is the 404 logic. It's spread out through a couple of places right now. It's in `datasette.utils.asgi.AsgiRouter`:

https://github.com/simonw/datasette/blob/227bb3e91fe34811a9374150798643a5af98ed79/datasette/utils/asgi.py#L98-L134

Then also in the `DatasetteRouter subclass of that:

https://github.com/simonw/datasette/blob/227bb3e91fe34811a9374150798643a5af98ed79/datasette/app.py#L736-L757

simonw:
[...]

PR #737, opened 2020-04-26
Custom pages mechanism, refs #648

simonw:
Refs #648. TODO:
- [x] Pass a `view_name` to `render_template()`
- [x] Mechanism for custom status code / headers / redirect
- [x] Documentation

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -762,10 +762,42 @@ class DatasetteRouter(AsgiRouter):
             except TemplateNotFound:
                 template = None
             if template:
-                await asgi_send_html(
+                headers = {}
+                status = [200]
+
+                def custom_header(name, value):
+                    headers[name] = value
+                    return ""
+
+                def custom_status(code):
+                    status[0] = code
+                    return ""
+
+                def custom_redirect(location, code=302):
+                    status[0] = code
+                    headers["Location"] = location
+                    return ""
+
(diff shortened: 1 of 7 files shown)


simonw:
> > Stretch goal: it would be neat if these pages could return custom HTTP headers (eg content-type) and maybe even status codes (eg for redirects) somehow.
> 
> I think I could do that with a custom template function - if that function is called during the render then we follow those instructions instead of returning the rendered HTML.
https://github.com/simonw/datasette/issues/648#issuecomment-619591380

simonw:
Design challenge: the function that can be called in the custom template to return a custom status code and/or headers.

Proposed design:
```
{{ custom_headers({"x-foo": "bar"}) }}
{{ custom_status_code(404) }}
```
Or should I combine them into one function?
```
{{ custom_headers({"x-foo": "bar"}, status=404) }}
```
A common use-case here would be a redirect. How should that look?
```
{{ custom_redirect("/foo") }} - defaults to 301
{{ custom_redirect("/foo", status=302) }}
```
Maybe avoid dictionary syntax entirely:
```
{{ custom_header("x-foo", "bar") }}
```

simonw:
I'm going with `custom_header(name, value)`, `custom_status(numeric_code)`, `custom_redirect("URL")`.

simonw:
Just needs documentation now.

Issue #731, opened 2020-04-22
Idea: option to automatically configure based on directory layout


simonw:
My Datasette projects increasingly take on the following structure:

- `metadata.json` with the metadata
- One or more `something.db` database files
- A `templates/` folder with some custom templates
- A `plugins/` folder with some custom plugins

Then I have to run Datasette like this:

    datasette *.db -m metadata.json --template-dir=templates --plugins-dir=plugins

It would be really interesting if Datasette had a special mode where you could point it at a directory with the above layout and it would automatically configure itself based on the contents.

Maybe even allow `datasette serve` to detect if it was passed a single argument that's a directory, not a file, and kick in to "directory layout configuration mode" in that case:

    datasette .

simonw:
I started thinking about this while building https://github.com/simonw/datasette-publish-now

The Vercel hosting platform is built on ASGI - so I need a way to provide an `app` ASGI object for it to use rather than Datasette's normal pattern of starting up the server using the `datasette serve` command.

The plugin currently does that by generating an `index.py` file that looks like this:

```python
from datasette.app import Datasette
import json
metadata = dict()
try:
    metadata = json.load(open("metadata.json"))
except Exception:
    pass
app = Datasette([], ["fixtures.db"], metadata=metadata).app()
```
But that means I have to replicate the different CLI options - `--plugins-dir` etc - and figure out how to pass them to the `Datasette()` constructor.

This would be a whole lot easier if I could use `app = Datasette(configure_directory=".")` and provide a magically configured instance based on the presence of files in the specified directory.

simonw:
I linked to this from https://github.com/zeit/now/discussions/4055

eyeseast:
I've almost suggested this same thing a couple times. I tend to have Makefile (because I'm doing other `make` stuff anyway to get data prepped), and I end up putting all those CLI options in something like `make run`. But it would be way easier to just have all those typical options -- plugins, templates, metadata -- be defaults.

simonw:
For `--static` this could work by letting you create the following:

- `static/css/`
- `static/js/`

Which would be automatically mounted at `/js/...` and `/css/...`

Or maybe just mount `static/` at `/static/` instead?

eyeseast:
Mounting `./static` at `/static` seems the simplest way. Saves you the trouble of deciding what else (`img` for example) gets special treatment.

simonw:
[...]

PR #739, opened 2020-04-26
Configuration directory mode

simonw:
Refs #731

TODO:

- [x] Decide how to combine explicit command-line options with items detected from the directory structure
- [x] Add unit tests
- [x] Implement `inspect-data.json` mechanism for populating `immutables`
- [x] Add documentation

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -194,21 +194,21 @@ class Datasette:
             self.add_database(db.name, db)
         self.cache_headers = cache_headers
         self.cors = cors
-        if (config_dir / "metadata.json").exists() and not metadata:
+        if config_dir and (config_dir / "metadata.json").exists() and not metadata:
             metadata = json.load((config_dir / "metadata.json").open())
         self._metadata = metadata or {}
         self.sqlite_functions = []
         self.sqlite_extensions = sqlite_extensions or []
-        if (config_dir / "templates").is_dir() and not template_dir:
+        if config_dir and (config_dir / "templates").is_dir() and not template_dir:
             template_dir = str((config_dir / "templates").resolve())
         self.template_dir = template_dir
-        if (config_dir / "plugins").is_dir() and not plugins_dir:
+        if config_dir and (config_dir / "plugins").is_dir() and not plugins_dir:
             template_dir = str((config_dir / "plugins").resolve())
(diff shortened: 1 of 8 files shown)


datasette/cli.py
```diff
@@ -352,21 +352,26 @@ def serve(
...
+            inspect_data=inspect_data,
+            metadata=metadata_data,
+            sqlite_extensions=sqlite_extensions,
+            template_dir=template_dir,
+            plugins_dir=plugins_dir,
+            static_mounts=static,
+            config=dict(config),
+            memory=memory,
+            version_note=version_note,
+        )
```
simonw, line 374: Design question: should I merge explicit options passed to `datasette serve` with automatically detected stuff in the directory? If so, what should over-ride what? or should it throw an error if you try to set something using both mechanisms?

simonw:
I think ALL explicit options should over-ride options that were detected in the directory structure. This means my current implementation isn't quite right - using a class method makes it hard to merge the details with the explicit options passed to the CLI function.

datasette/cli.py
simonw: I think this logic should live in the `Datasette()` constructor. But the special case that says "if files is a single item that's a directory" should live in the CLI code.

So the directory option would be something like:

     Datasette(config_dir=".")

simonw:
I'm going to add a `Datasette(config_dir=".")` parameter which initializes from the given directory, allowing other parameters to over-ride it.

datasette/app.py
```diff
@@ -164,9 +164,16 @@ def __init__(
         memory=False,
         config=None,
         version_note=None,
+        config_dir=None,
     ):
+        assert config_dir is None or isinstance(
+            config_dir, Path
+        ), "config_dir= should be a pathlib.Path"
+        # TODO: Use 'inspect-data.json' to decide on immutables
         immutables = immutables or []
```
simonw, line 173: If `inspect-data.json` is present and a database file is listed there, treat that file as immutable.

simonw:
I think documentation goes on https://datasette.readthedocs.io/en/stable/config.html - but I'm going to need to re-arrange that page to fit in a section about configuration directories.

I'll move the existing list of settings into a "configration options" section.

PR #796, opened 2020-06-03
New WIP writable canned queries

simonw:
Refs #698. Replaces #703

Still todo:

- [x] Unit tests
- ~~Figure out `.json` mode~~
- [x] Flash message solution
- ~~CSRF protection~~
- [x] Better error message display on errors
- [x] Documentation
- ~~Maybe widgets?~~ I'll do these later

### DIFF
--- a/datasette/templates/query.html
+++ b/datasette/templates/query.html
@@ -27,11 +27,12 @@
 {% endblock %}
 
 {% block content %}
+
 <h1 style="padding-left: 10px; border-left: 10px solid #{{ database_color(database) }}">{{ metadata.title or database }}</h1>
 
 {% block description_source_license %}{% include "_description_source_license.html" %}{% endblock %}
 
-<form class="sql" action="{{ database_url(database) }}{% if canned_query %}/{{ canned_query }}{% endif %}" method="get">
+<form class="sql" action="{{ database_url(database) }}{% if canned_query %}/{{ canned_query }}{% endif %}" method="{% if canned_write %}post{% else %}get{% endif %}">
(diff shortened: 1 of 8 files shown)


simonw:
"Query executed" is the default message, but it's pretty bland:

data__insert_into_names__name__values___name_

How about letting queries define custom success messages in their metadata configuration?

simonw:
For `.json` mode (when the URL has a `.json` suffix) I'm going to handle both form-encoded and JSON encoded inputs and return a 200 status code with JSON that tells you if the query executed successfully and how many rows were affected.

simonw:
Default message is now "Query executed, 1 row affected" as of https://github.com/simonw/datasette/pull/796/commits/f45c44ac8f3a9a2182d76c6bda44a06676499e4b

simonw:
I'm going to have `on_success_url` and `on_error_url` properties for specifying where the redirect should go too.

simonw:
I'm having trouble coming up with a canned SQL query that will only error with certain parameters, which I need in order to test out the `on_error_redirect` and `on_error_message` properties.

simonw:
Maybe I need some kind of optional validation mechanism? SQLite lets me insert floating point numbers into integer columns right now, which feels wrong.

simonw:
Violating a unique constraint will throw an error.

simonw:
Consider this one:
```
                        "delete_name": {
                            "sql": "delete from names where rowid = :rowid",
                            "write": True,
                            "on_success_message": "Name deleted",
                        },
```
If the user enters an invalid `rowid` the query will still execute without errors and hence the success message will still be displayed. Can I address this?

Maybe allow an optional `"rowcount_expected": 1` property?

simonw:
I'm going to document this, land it and then continue to work on the other pieces - CSRF protection and .json mode - in separate tickets.

Issue #793, opened 2020-06-02
CSRF protection for /-/messages tool and writable canned queries


simonw:
> The `/-/messages` debug tool will need CSRF protection or people will be able to add messages using a hidden form on another website.
_Originally posted by @simonw in https://github.com/simonw/datasette/issues/790#issuecomment-637790860_

simonw:
This is a minor security issue with `master` at the moment, but I'll resolve this before I ship the next release.

simonw:
I need this for writable canned queries in #796 too.

simonw:
I need to land and release the fix for signing cookies in https://github.com/simonw/asgi-csrf/issues/2

PR #798, opened 2020-06-05
CSRF protection

simonw:
Refs #793

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1,4 +1,5 @@
 import asyncio
+import asgi_csrf
 import collections
 import datetime
 import hashlib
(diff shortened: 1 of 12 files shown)


simonw:
Needs unit tests.

More importantly: needs very, very careful consideration of how this plays with HTTP caching.

simonw:
I don't want to set a cookie on a page response that is being cached.

Right now the ASGI middleware will be doing exactly that, which is bad.

But how do I get nature that when you load a page with a form that will be CSRF protected you have been served the cookie?

Maybe those pages should do something explicit to the request object indicating that the cookie is needed?

That works for Datasette (since it has mutable request objects) but I'm not sure how it would work in the asgi-csrf pure ASGI middleware context.

simonw:
Django docs on CSRF and caching: https://docs.djangoproject.com/en/3.0/ref/csrf/#caching

simonw:
I'm solving the compatibility with caching problem in this ticket: https://github.com/simonw/asgi-csrf/issues/7

simonw:
Add unit tests illustrating the `Vary: Cookie` header and I'm done here.

PR #557, opened 2019-07-11
Get tests running on Windows using Travis CI

simonw:
Refs #511

PR #868, opened 2020-06-26
initial windows ci setup

joshmgrant:
Picking up the work done on #557 with a new PR. Seeing if I can get this working.


simonw:
This is really exciting! Thanks so much for looking into this.

I'm interested in moving CI for this repo over to GitHub Actions, so I'd be fine with you getting this to work as an Action rather than through Travis. If you can get it working in Travis though I'll happily land that and figure out how to convert that to GitHub Actions later on.

joshmgrant:
Ok, so it's been a while but I think I'm making progress. The good news: I have come up with a configuration change for running the tests on Windows on GitHub Actions. The bad news: there's a bunch of extraneous errors on the Windows case. I *think* this is due to Windows file IO and sqlite in a lot of cases, so I'm working through it. I will eventually clean up this PR.

PR #890, opened 2020-07-03
Load only python files from plugins-dir.

amjith:
The current behavior for `--plugins-dir` is to load every file in that folder as a python module. This can result in errors if there are non-python files in the plugins dir (such as .mypy_cache). 

This PR restricts the module loading to only python files.


simonw:
This is an excellent fix. Thanks!

Not sure why codecov is complaining. I'm going to merge it as-is.

PR #902, opened 2020-07-21
Don't install tests package

abeyerpath:
The `exclude` argument to `find_packages` needs an iterable of package
names.

Fixes: #456


simonw:
Thanks for spotting this!

PR #986, opened 2020-10-01
Allow facet by primary keys, fixes #985

### LINKED ISSUE #985: Column actions should support facet by compound primary keys
On https://latest.datasette.io/fixtures/compound_three_primary_keys the column action menu doesn't display for the pk1, pk2 and pk3 columns (because they are primary keys) even though faceting by them is actually useful.

Refs #98

MrNaif2018:
Hello! This PR makes it possible to facet by primary keys.
Did I get it right that just removing the condition on UI side is enough? From testing it works fine with primary keys, just as with normal keys.
If so, should I also remove unused `data-is-pk`?


simonw:
Thanks for taking a look! The fix ended up being a little different from this because I still want to disable faceting on regular single primary keys (since faceting by those won't ever produce interesting results) - here's what I used: https://github.com/simonw/datasette/commit/5d6bc4c268f9f155e59561671f8617addd3e91bc

PR #995, opened 2020-10-06
Document setting Google Cloud SDK properties

ghing:
Document setting Google Cloud SDK properties to avoid having to respond to interactive prompts when running `datasette publish cloudrun`.


simonw:
This is great, thank you!

PR #1022, opened 2020-10-14
Fix table name in spatialite example command

jsfenfen:
The example query for creating a new point geometry seems to be using a table called 'museums' but at one point it instead uses 'events'. I *believe* it is intended to be museums (the example makes more sense if so).


simonw:
Good catch, thanks!

PR #1030, opened 2020-10-18
Make `package` command deal with a configuration directory argument

frankier:
Currently if we run `datasette package` on a configuration directory we'll get an exception when we try to hard link to the directory. This PR copies the tree and makes the Dockerfile run inspect on all *.db files.


frankier:
Not currently using this myself and it looks like this code has been reworked, so I assume this has been fixed no.w

PR #1031, opened 2020-10-19
Fallback to databases in inspect-data.json when no -i options are passed

frankier:
Currenlty `Datasette.__init__` checks immutables against None to decide whether to fallback to inspect-data.json. This patch modifies the serve command to pass None when no -i options are passed so this fallback works correctly.

### DIFF
--- a/datasette/cli.py
+++ b/datasette/cli.py
@@ -412,7 +412,7 @@ def serve(
         metadata_data = parse_metadata(metadata.read())
 
     kwargs = dict(
-        immutables=immutable,
+        immutables=immutable or None,
         cache_headers=not reload,
         cors=cors,
         inspect_data=inspect_data,


simonw:
I don't fully understand the bug you're fixing here. Could you provide a bit more explanation?

frankier:
The bug is that currently when there are databases passed in, but no -i flag, e.g. in configuration directory mode, inclusion in inspect-data.json does not automatically cause databases to be considered immutable, as described in the documentation.

The reason is that the -i flag is specified multiple=True, which means when it is not passed in we will get an empty list [], rather than None. So the current code decides that no databases are immutable rather than falling back to inspect-data.json -- as is presumably intended.

frankier:
Please let me know if there's anything I can do to help get this merged.

This is causing problems for me because it means when I build my Docker image my databases aren't considered immutable, which I would like them to be so that a download link is produced.

simonw:
This bug should have been fixed in #1229 - let me know if that's not the case!

Thanks

simonw:
Sorry I didn't get to this PR sooner. I've joint-credited you in the release notes for this fix: https://docs.datasette.io/en/stable/changelog.html#v0-56

PR #1038, opened 2020-10-21
DOC: Fix syntax error

gerrymanoim:
If I understand https://docs.datasette.io/en/stable/plugin_hooks.html#register-routes correctly, `register_routes` should return a `List[Tuple[str, Callable]]`. I believe the current code in documentation has a syntax error (extra `)`).


simonw:
Thanks for spotting this!

PR #1044, opened 2020-10-23
Add minimum supported python

bollwyvl:
Thanks for `datasette`!

This PR adds `python_requires` to formally signal the [minimum supported python version](https://packaging.python.org/guides/dropping-older-python-versions/#specify-the-version-ranges-for-supported-python-distributions) (which is pointed out with classifiers, so seems pretty straightforward).


simonw:
Thanks for this!

Issue #1042, opened 2020-10-23
Plugin hook for loading templates


simonw:
This can work with the Jinja template loaders. It would unlock things like storing templates in SQLite.

simonw:
Relevant code: https://github.com/simonw/datasette/blob/d0cc6f4c32e1f89238ddec782086b3122f445bd4/datasette/app.py#L288-L311

simonw:
Maybe `register_template_loader()` which returns an object which is added in at the beginning of the list passed to `ChoiceLoader` here.

simonw:
I'll build this in conjunction with a plugin that supports editing templates stored in SQLite.

simonw:
A better version of this hook would be passed the database, table and query name depending on what was being rendered.

This would require some re-thinking of how core templates are loaded, especially since I would want the templates considered comment to continue working.

simonw:
When loading a template the filename is required, but you can optionally also send a set of extra arguments which the template loader can take into consideration.

simonw:
Maybe the template loader can optionally return some extra context to pass to the template. That could be used to solve the templates considered comment.

simonw:
Alternative plugin hook idea:
```python
@hookspec
def load_template(template, database, table, columns, view_name, request, datasette):
    "Load a template"
```
Imitating the existing `extra_template_vars` family of hooks: https://docs.datasette.io/en/stable/plugin_hooks.html#extra-template-vars-template-database-table-columns-view-name-request-datasette

simonw:
Almost all of the core template loading happens in the `BaseView.render` method: https://github.com/simonw/datasette/blob/976e5f74aae1fa0d406df6691dc8b5feeebe8788/datasette/views/base.py#L114-L133

The one exception is the 404 handling code here: https://github.com/simonw/datasette/blob/976e5f74aae1fa0d406df6691dc8b5feeebe8788/datasette/app.py#L1034-L1042

simonw:
Also consider that `DatasetteRouter` uses `.list_templates()` to gather together `{slug}.html` style templates for the custom page templates mechanism: https://github.com/simonw/datasette/blob/976e5f74aae1fa0d406df6691dc8b5feeebe8788/datasette/app.py#L949-L967

For that to work with the new plugin hook, custom template providing plugins will need a way to provide a list of templates that they know about.

simonw:
[...]

PR #1069, opened 2020-10-30
load_template() plugin hook

simonw:
Refs #1042

### DIFF
--- a/datasette/hookspecs.py
+++ b/datasette/hookspecs.py
@@ -50,6 +50,11 @@ def extra_template_vars(
 
 
 @hookspec
+def load_template(template, database, table, columns, view_name, request, datasette):
+    "Load the specified template, returning the template code as a string"
+
+
+@hookspec
 def publish_subcommand(publish):
     "Subcommands for 'datasette publish'"
 
--- a/datasette/views/base.py
+++ b/datasette/views/base.py
@@ -148,10 +148,8 @@ class BaseView:
             **context,
             **{
                 "database_color": self.database_color,
-                "templates_considered": [{
-                    "name": template.name,
-                    "used": template_name == template.name
-                }
+                "templates_considered": [
+                    {"name": template.name, "used": template_name == template.name}
                     for template_name in templates
                 ],
             },
--- a/datasette/views/base.py
+++ b/datasette/views/base.py
@@ -119,15 +119,39 @@ class BaseView:
 
     async def render(self, templates, request, context=None):
         context = context or {}
-        template = self.ds.jinja_env.select_template(templates)
+
+        # Give plugins first chance at loading the template
+        break_outer = False
+        plugin_template_source = None
+        template_name = None
+        for template_name in templates:
+            if break_outer:
+                break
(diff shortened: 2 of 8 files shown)


simonw:
Next steps: build a demonstration plugin against this.

datasette/views/base.py
```diff
@@ -119,15 +119,37 @@ async def dispatch_request(self, request, *args, **kwargs):
...
         template_context = {
             **context,
             **{
                 "database_color": self.database_color,
-                "select_templates": [
-                    "{}{}".format(
-                        "*" if template_name == template.name else "", template_name
-                    )
+                "templates_considered": [
+                    {"name": template.name, "used": template_name == template.name}
```
simonw, line 152: I'd like to include additional information in this debug output telling you if the template was returned from a plugin.

datasette/hookspecs.py
```diff
@@ -49,6 +49,11 @@ def extra_template_vars(
     "Extra template variables to be made available to the template - can return dict or callable or awaitable"
 
 
+@hookspec
+def load_template(template, database, table, columns, view_name, request, datasette):
```
simonw, line 53: I think this should use `firstresult=True`: https://docs.pytest.org/en/latest/writing_plugins.html#firstresult

datasette/views/base.py
simonw: Unfortunately I don't think Pluggy supports that at the moment. There's a feature request for it here: https://github.com/pytest-dev/pluggy/issues/235

simonw: I could at least indicate if the template was loaded from the filesystem or was provided by a plugin. That should be enough information for most cases.

simonw:
I should include an example in the documentation that shows loading templates from a database table.

simonw:
Bringing over a comment from #1042:

> I'd like to do this all in the `datasette.render_template()` method to ensure it's available to plugins as well, not just core code that uses the `BaseView` class.
> 
> This code is the problem:
> 
> https://github.com/simonw/datasette/blob/d3e9b0aecb6f8e9b2befd9c654ccb7ce852db3e7/datasette/views/base.py#L114-L133
> 
> I think I'll fix this by moving the `select_templates` mechanism into `datasette.render_templates()`.

simonw:
I'm having second thoughts about the design of the plugin hook.

Consider the following:
```python
            plugin_template_source = pm.hook.load_template(
                template=template_name,
                database=context.get("database"),
                table=context.get("table"),
                columns=context.get("columns"),
                view_name=self.name,
                request=request,
                datasette=self.ds,
            )
```
It's a bit gross that `database`, `table` and `columns` are pulled out of the context like that. This doesn't make sense for pages that are rendered by plugins, for example.

So maybe for the first release of this plugin hook I should cut it down to just seeing `template`, `request` and `datasette`. I can add the table/view/etc stuff back in later if it turns out to be necessary.

simonw:
OK, new hook specification is:
```python
@hookspec(firstresult=True)
def load_template(template, request, datasette):
    "Load the specified template, returning the template code as a string"
```

PR #1112, opened 2020-11-28
Fix --metadata doc usage

jefftriplett:
I stumbled on this while trying to figure out how to configure datasette-ripgrep via https://github.com/simonw/datasette-ripgrep/issues/15

You may not want to update the changelog (those are annoying) so I added two commits in case that's easier.


jefftriplett:
(Even more annoying is that I see my editor leaked an extra delete space at the end of the line. I'm happy to rebuild this to be less annoying, but you probably don't want the changelog update either way)

simonw:
Thanks!

Issue #1116, opened 2020-11-30
GENERATED column support


nattaylor:
I think this is a feature request... perhaps I should just try to contribute it myself, but thought I'd check in case support is planned already.

For a table with the following schema, datasette 0.51.1 doesn't pick up the GENERATED columns and the column list only contains `(rowid, body)`  If I edit the SQL and select the generated columns, it will happily show them.

At first glance it appears that [`def table_column_details(conn, table):`](https://github.com/simonw/datasette/blob/4777362bf2692bc72b221ec47c3e6216151d1b89/datasette/utils/__init__.py#L575) would have to be refactored to use a different methodology to get the columns, since `PRAGMA table_info(deeds);` returns just `0|body|TEXT|0||0` so maybe it wouldn't be worth it.

```
CREATE TABLE deeds (
    body TEXT,
    id INT GENERATED ALWAYS AS (json_extract(body, '$.id')) STORED,
    consideration INT GENERATED ALWAYS AS (json_extract(body, '$.consideration')) STORED
);
```

simonw:
I'm treating this as a bug - these columns should definitely be visible in Datasette.

I created my own test database using SQLite from Homebrew like this:
```
/usr/local/Cellar/sqlite/3.33.0/bin/sqlite3 deeds.db << EOF
CREATE TABLE deeds (
    body TEXT,
    id INT GENERATED ALWAYS AS (json_extract(body, '$.id')) STORED,
    consideration INT GENERATED ALWAYS AS (json_extract(body, '$.consideration')) STORED
);
INSERT INTO deeds (body) VALUES ('{
    "id": 1,
    "consideration": "This is the consideration"
}');
EOF
```

simonw:
It looks like `PRAGMA table_info` skips "hidden" columns: https://www.sqlite.org/pragma.html#pragma_table_info

But `PRAGMA table_xinfo` does not: https://www.sqlite.org/pragma.html#pragma_table_xinfo

nattaylor:
I poked around a little more.  The columns are included in `table_xinfo()` but this also includes other columns that might not want to be exposed.

```
sqlite> PRAGMA table_xinfo(deeds);
0|body|TEXT|0||0|0
1|id|INT GENERATED ALWAYS|0||0|3
2|consideration|INT GENERATED ALWAYS|0||0|3
```

simonw:
Generated column support was added in SQLite 3.31.0, so any unit tests I write for this should use skipIf to only run on that version or later.

[...]

PR #1117, opened 2020-11-30
Support for generated columns

simonw:
Refs #1116. My first attempt at this worked on my laptop but broke in CI, so I'm going to iterate on it in a pull request instead.

### DIFF
--- a/datasette/utils/sqlite.py
+++ b/datasette/utils/sqlite.py
@@ -1,0 +1,19 @@
+try:
+    import pysqlite3 as sqlite3
+except ImportError:
+    import sqlite3
+
+if hasattr(sqlite3, "enable_callback_tracebacks"):
+    sqlite3.enable_callback_tracebacks(True)
+
+
+def sqlite_version():
+    return tuple(
+        map(
+            int,
+            sqlite3.connect(":memory:")
+            .execute("select sqlite_version()")
+            .fetchone()[0]
+            .split("."),
+        )
+    )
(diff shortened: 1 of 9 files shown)


simonw:
I need to replicate these failures on my laptop. My hunch is that this is down to the version of SQLite available to Python.

simonw:
On my laptop:
```
platform darwin -- Python 3.8.6, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
SQLite: 3.33.0
```
In CI they are all SQLite: 3.22.0

simonw:
This kind of problem is why I have a `tmate` workflow:

Actions_·_simonw_datasette

simonw:
Here's the problem: https://www.sqlite.org/changes.html#version_3_26_0

> ### 2018-12-01 (3.26.0)
> 
> -  Added [PRAGMA table_xinfo](https://www.sqlite.org/pragma.html#pragma_table_xinfo) that works just like [PRAGMA table_info](https://www.sqlite.org/pragma.html#pragma_table_info) except that it also shows [hidden columns](https://www.sqlite.org/vtab.html#hiddencol) in virtual tables.

CI is running 3.22.0.

datasette/utils/sqlite.py
```diff
@@ -0,0 +1,19 @@
+try:
+    import pysqlite3 as sqlite3
+except ImportError:
+    import sqlite3
+
+if hasattr(sqlite3, "enable_callback_tracebacks"):
+    sqlite3.enable_callback_tracebacks(True)
+
+
+def sqlite_version():
```
simonw, line 10: I'm nervous about the performance impact of this, since every time it is called it creates a new in-memory connection.

I'm going to cache this so that it only ever gets called a maximum of once for the lifetime of the server.

nattaylor:
I just deployed this and its working great.

In a very unscientific benchmark my response times went from around 22-25ms to 33-36ms, but I didn't even dig enough to confirm the latency is related to the change.  It's on a VPS, so maybe the load changed.

Issue #1121, opened 2020-12-01
Table actions cog is misaligned


abdusco:
At the moment it looks like this
https://datasette-graphql-demo.datasette.io/github/repos

![image](https://user-images.githubusercontent.com/3243482/100716533-e6e2d300-33c9-11eb-866e-1e83ba228bf5.png)

Adding a few flex statements fixes the alignment and centers `h1` text and the cog icon vertically.
![image](https://user-images.githubusercontent.com/3243482/100716605-f8c47600-33c9-11eb-8d69-0e37499cf641.png)

simonw:
Demo: https://latest.datasette.io/fixtures?_bot=1

<img width="729" alt="fixtures" src="https://user-images.githubusercontent.com/9599/100950004-43c0b500-34c0-11eb-918c-aa959376461f.png">

PR #1122, opened 2020-12-01
Fix misaligned table actions cog

abdusco:
Fixes https://github.com/simonw/datasette/issues/1121


simonw:
This is a very neat fix, thank you.

Issue #1094, opened 2020-11-14
import EX_CANTCREAT means datasette fails to work on Windows


drkane:
Trying to use datasette 0.51.1 gives the following error:

```
ImportError: cannot import name 'EX_CANTCREAT' from 'os' (C:\Users\drkan\AppData\Local\Programs\Python\Python39\lib\os.py)
```

Looks like that code is only available on unix: https://docs.python.org/3/library/os.html#os.EX_CANTCREAT

Removing the line makes it work fine (`EX_CANTCREAT` doesn't seem to be used anywhere?)

bapowell:
I can confirm this issue, running version 0.51.1 under Windows.

Fixed by commenting out the following line near the top of datasette\utils\asgi.py :

`#from os import EX_CANTCREAT`

PR #1128, opened 2020-12-04
Fix startup error on windows

abdusco:
Fixes https://github.com/simonw/datasette/issues/1094

This import isn't used at all, and causes error on startup on Windows.


simonw:
Thanks for this!

PR #1158, opened 2020-12-23
Modernize code to Python 3.6+

eumiro:
- compact dict and set building
- remove redundant parentheses
- simplify chained conditions
- change method name to lowercase
- use triple double quotes for docstrings

please feel free to accept/reject any of these independent commits

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -429,7 +429,7 @@ class Datasette:
             return m
 
     def plugin_config(self, plugin_name, database=None, table=None, fallback=True):
-        "Return config for plugin, falling back from specified database/table"
+        """Return config for plugin, falling back from specified database/table"""
         plugins = self.metadata(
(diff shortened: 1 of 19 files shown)


simonw:
Did you use a tool to find these or are they all from manual code inspection?

They look good - I particularly like the set/dict  literals.

eumiro:
The dict/set suggestion comes from `pyupgrade --py36-plus`, but then had to `black` the change.

The rest comes from PyCharm's Inspect code function. I reviewed all the suggestions and fixed a thing or two, such as leading/trailing spaces in the docstrings or turned around the chained conditions.

Then I tried to convert all `os.path/glob/open` to `Path`, but there were some local test issues, so I'll have to start over in smaller chunks if you want to have that too.

simonw:
Thanks for this!

I'm fine keeping the `os.path` stuff as is.

Issue #1191, opened 2021-01-15
Ability for plugins to collaborate when adding extra HTML to blocks in default templates


simonw:
Sometimes a plugin may want to add content to an existing default template - for example `datasette-search-all` adds a new search box at the top of `index.html`. I also want `datasette-upload-csvs` to add a CTA on the `database.html` page: https://github.com/simonw/datasette-upload-csvs/issues/18

Currently plugins can do this by providing a new version of the `index.html` template - but if multiple plugins try to do that only one of them will succeed.

It would be better if there were known areas of those templates which plugins could add additional content to, such that multiple plugins can use the same spot.

simonw:
This relates to #987 (documented HTML hooks for JavaScript plugins) but is not quite the same thing.

simonw:
Also related: #857 (comprehensive documentation of variables available to templates) - since then the plugin hook could be fed the full template context and use that to do its thing.

Or maybe the plugin hooks get to return the name of a template that should be `{% include %}` into the page at that point? But the plugin may want to add extra context that is available to that template include.

simonw:
Plugins that want to provide extra context to the template can already do so using the `extra_template_vars()` plugin hook.

So this hook could work by returning a list of template filenames to be included. Those templates can be bundled with the plugin. Since they are included they will have access to the template context and to any `extra_template_vars()` values.

simonw:
I think this ends up being a whole collection of new plugin hooks, something like:

- `include_table_top`
- `include_table_bottom`
- `include_row_top`
- `include_row_bottom`
- `include_database_top`
- `include_database_bottom`
- `include_query_bottom`
- `include_query_bottom`

simonw:
I'm going to prototype this in a branch.

simonw:
I'm going to try using Jinja macros to implement this: https://jinja.palletsprojects.com/en/2.11.x/templates/#macros

simonw:
Another potential use for this: plugins that provide authentication (like `datasette-auth-passwords` and `datasette-auth-github`) could use it to add a chunk of HTML to the "permission denied" page that links to their mechanism of authenticating.

simonw:
I can combine this with #987 - each of these areas of the page can be wrapped in a `<div>` with a class that matches the name of the plugin hook, that way JavaScript plugins can append their content in the same place as Python plugins.

simonw:
[...]

PR #1204, opened 2021-01-25
WIP: Plugin includes

simonw:
Refs #1191

Next steps:

- [ ] Get comfortable that this pattern is the right way to go
- [ ] Implement it for all of the other pages, not just the table page
- [ ] Add a new set of plugin tests that exercise ALL of these new hook locations
- [ ] Document, then ship


simonw:
Writing the tests will be a bit tricky since we need to confirm that the `include_table_top(datasette, database, actor, table)` arguments were all passed correctly but the only thing we get back from the plugin is a list of templates. Maybe encode those values into the template names somehow?

simonw:
I think this is missing the `_macro.html` template file but I have that in my Dropbox.

Issue #1208, opened 2021-01-26
A lot of open(file) functions are used without a context manager thus producing ResourceWarning: unclosed file <_io.TextIOWrapper


kbaikov:
Your code is full of open files that are never closed, especially when you deal with reading/writing json/yaml files.

If you run python with warnings enabled this problem becomes evident.
This probably contributes to some memory leaks in long running datasettes if the GC will not 'collect' those resources properly.

This is easily fixed by using a context manager instead of just using open:
```python
with open('some_file', 'w') as opened_file:
    opened_file.write('string')
```

In some newer parts of the code you use Path objects 'read_text' and 'write_text' functions which close the file properly and are prefered in some cases.

If you want I can create a PR for all places i found this pattern in.

Bellow is a fraction of places where i found a ResourceWarning:
```python

update-docs-help.py:
  20          actual = actual.replace("Usage: cli ", "Usage: datasette ")
  21:         open(docs_path / filename, "w").write(actual)
  22  

datasette\app.py:
  210          ):
  211:             inspect_data = json.load((config_dir / "inspect-data.json").open())
  212              if immutables is None:

  266          if config_dir and (config_dir / "settings.json").exists() and not config:
  267:             config = json.load((config_dir / "settings.json").open())
  268          self._settings = dict(DEFAULT_SETTINGS, **(config or {}))

  445              self._app_css_hash = hashlib.sha1(
  446:                 open(os.path.join(str(app_root), "datasette/static/app.css"))
  447                  .read()

datasette\cli.py:
  130      else:
  131:         out = open(inspect_file, "w")
  132      loop = asyncio.get_event_loop()

  459      if inspect_file:
  460:         inspect_data = json.load(open(inspect_file))
  461  

```

simonw:
This is a good catch - I've been lazy about this, but you're right that it's an issue that needs cleaning up.  Would be very happy to apply a PR, thanks!

kbaikov:
@simonw could you please take a look at the PR 1211 that fixes this issue?

PR #1211, opened 2021-01-31
Use context manager instead of plain open

kbaikov:
Context manager with open closes the files after usage. Fixes: https://github.com/simonw/datasette/issues/1208

When the object is already a pathlib.Path i used read_text
write_text functions

In some cases pathlib.Path.open were used in context manager,
it is basically the same as builtin open.

Tests are passing: 850 passed, 5 xfailed, 10 xpassed

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -208,7 +208,7 @@ class Datasette:
             and (config_dir / "inspect-data.json").exists()
             and not inspect_data
         ):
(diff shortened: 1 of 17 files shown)


kbaikov:
Ping @simonw

simonw:
Thanks very much for this - it's really comprehensive. I need to bake some of these patterns into my coding habits better!

PR #1229, opened 2021-02-17
ensure immutable databses when starting in configuration directory mode with

### LINKED ISSUE #1224: can't start immutable databases from configuration dir mode
Say I have a `/databases/` directory with multiple sqlite db files in that dir (`1.db` & `2.db`) and an `inspect-data.json` file.

If I start datasette via `datasette -h 0.0.0.0 /databases/` then the resulting databases are set to `is_mutable: true` as inspected via http://127.0.0.1:8001/-/databases.json

I don't want to have to list out the databases by name, e.g. `datasette -i /databases/1.db -i /databases/2.db` as i want the system to autodetect the sqlite dbs i have in the configuration directory 

According to the docs outlined in https://docs.datasette.io/en/latest/settings.html?highlight=immutable#configuration-directory-mode this should be possible
> `inspect-data.json` the result of running datasette inspect - any database files listed here will be treated as immutable, so they should not be changed while Datasette is running
 
I believe that if the `inspect-json.json` file present, then in theory the databases will be automatically set to immutable via this code https://github.com/simonw/datasette/blob/9603d893b9b72653895318c9104d754229fdb146/datasette/app.py#L211-L216

However it appears the Click Multiple Options will return a tuple via https://github.com/simonw/datasette/blob/9603d893b9b72653895318c9104d754229fdb146/datasette/cli.py#L311-L317

The resulting tuple is passed to the Datasette app via `kwargs` and overrides the behaviour to set the databases to immutable via this arg https://github.com/simonw/datasette/blob/9603d893b9b72653895318c9104d754229fdb146/datasette/app.py#L182

If you think this is a bug and needs fixing, I am willing to make a PR to check for the empty `immutable` tuple before calling the Datasette class initializer as I think leaving that class interface alone is the best path here.

Thoughts?

Also - i'm loving Datasette, it truly is a wonderful tool, thank you :)

camallen:
fixes #1224 

This PR ensures all databases found in a configuration directory that match the files in `inspect-data.json` will be set to `immutable` as outlined in https://docs.datasette.io/en/latest/settings.html#configuration-directory-mode

specifically on building the `datasette` instance it checks:
- if `immutables` is an empty tuple - as passed by the cli code
- if `immutables` is the default function value `None` - when it's not explicitly set

And correctly builds the immutable database list from the `inspect-data[file]` keys.

Note for this to work the `inspect-data.json` file must contain `file` paths which are relative to the configuration directory otherwise the file paths won't match and the dbs won't be set to immutable. 

I couldn't find an easy way to test this due to the way `make_app_client` works, happy to take directions on adding a test for this. 

I've updated the relevant docs as well, i.e. use the `inspect` cli cmd from the config directory path to create the relevant file
```
cd $config_dir
datasette inspect *.db --inspect-file=inspect-data.json
```
https://docs.datasette.io/en/latest/performance.html#using-datasette-inspect


camallen:
I believe this pr and #1031 are related and fix the same issue.

simonw:
Thanks for figuring this out!

PR #1252, opened 2021-03-09
Add back styling to lists within table cells (fixes #1141)

### LINKED ISSUE #1141: Default styling for bullet point lists
I just noticed that https://datasette.io/content/recent_releases (which uses `datasette-render-markdown`) is missing its bullet points:

content__recent_releases__399_rows

bobwhitelock:
This overrides the Datasette reset - see https://github.com/simonw/datasette/blob/d0fd833b8cdd97e1b91d0f97a69b494895d82bee/datasette/static/app.css#L35-L38 - to add back the default styling of list items displayed within Datasette table cells.

Following this change, the same content as in the original issue looks like this:

### DIFF
--- a/datasette/static/app.css
+++ b/datasette/static/app.css
@@ -452,6 +452,10 @@ table a:link {
     margin-left: -10%;
     font-size: 0.8em;
 }
+.rows-and-columns td ol,ul {
+    list-style: initial;
+    list-style-position: inside;
+}
 a.blob-download {
     display: inline-block;
 }


datasette/static/app.css
```diff
@@ -452,6 +452,10 @@ table a:link {
     margin-left: -10%;
     font-size: 0.8em;
 }
+.rows-and-columns td ol,ul {
+    list-style: initial;
+    list-style-position: inside;
```
bobwhitelock, line 457: This is needed otherwise the bullet point marker appears in the neighbouring cell to the left, rather than the cell with the bullet point content.

simonw:
Neat fix, thank you!

Issue #1270, opened 2021-03-22
Try implementing SQLite timeouts using .interrupt() instead of using .set_progress_handler()


simonw:
> Maybe I could implement SQLite query timeouts using the `interrupt()` method instead of the progress handler hack I'm currently using?
>
> https://stackoverflow.com/questions/43240496/python-sqlite3-how-to-quickly-and-cleanly-interrupt-long-running-query-with-e has some tips.

_Originally posted by @simonw in https://github.com/simonw/datasette/issues/1268#issuecomment-803764919_

simonw:
I think the implementation fir this goes here: https://github.com/simonw/datasette/blob/6f41c8a2bef309a66588b2875c3e24d26adb4850/datasette/database.py#L146-L157

I figured out a similar pattern in `datasette-ripgrep` here: https://github.com/simonw/datasette-ripgrep/blob/0.7/datasette_ripgrep/__init__.py#L63-L71

simonw:
Confirmed that the interrupt() based cancellation mechanism fixes the SpatiaLite issue in #1268!

simonw:
I managed to build SpatiaLite such that this isn't necessary any more. I'm still interested in pursuing this further though - it feels like it could be a more robust way of implementing timeouts, but I need to prove to myself that it's better (maybe better performance, or handles more edge-cases?). Not sure how to prove that yet.

Issue #1268, opened 2021-03-22
Figure out why SpatiaLite 5.0 hangs the database page on Linux


simonw:
See detailed notes in https://github.com/simonw/datasette/issues/1249 - for some reason SpatiaLite 5.0 hangs the `/dbname` page on Linux (inside Docker containers, both with a custom compiled SpatiaLite and one installed from the Ubuntu 20.10 package repository). This doesn't happen on macOS with SpatiaLite 5 installed using Homebrew.

simonw:
`gdb` may be able to help debug this: https://www.podoliaka.org/2016/04/10/debugging-cpython-gdb/

simonw:
Here's a fun debugging trick:

    docker run -it -p 8001:8001 -v `pwd`:/mnt datasette-spatialite:latest bash
    root@16197781a7b5:/# python3 -m trace --trace $(which datasette) \
      -p 8001 -h 0.0.0.0 /mnt/tuscany_housenumbers.sqlite \
      --load-extension=spatialite

A huge amount of stuff scrolls past as Datasette starts up, since we are tracing every executed line of Python.

After about a minute it's finished starting and I can make some HTTP requests against it.

simonw:
Hitting http://localhost:8001/ successfully shows the homepage (after a lot more scrolling).

simonw:
Hitting http://localhost:8001/tuscany_housenumbers triggers the bug. It gets stuck in a loop that looks like this:

<img width="381" alt="datasette_—_root_16197781a7b5____—_com_docker_cli_◂_docker_run_-it_-p_8001_8001_-v___Dropbox_Development_datasette__mnt_datasette-spatialite_latest_bash_—_195×48_and_getIncidentsGit_—_-zsh_—_162×60" src="https://user-images.githubusercontent.com/9599/111942115-e0b75600-8a8f-11eb-8b6d-5c37240ce2f8.png">

Which looks to me like this code: https://github.com/simonw/datasette/blob/8e18c7943181f228ce5ebcea48deb59ce50bee1f/datasette/utils/__init__.py#L139-L158

simonw:
So I think there's a bug in the way the `set_progress_handler()` mechanism works when used in conjunction with SpatiaLite 5.0 on Linux.

simonw:
[...]

PR #1271, opened 2021-03-22
Use SQLite conn.interrupt() instead of sqlite_timelimit()

simonw:
Refs #1270, #1268, #1249

Before merging this I need to do some more testing (to make sure that expensive queries really are properly cancelled). I also need to delete a bunch of code relating to the old mechanism of cancelling queries.

[See comment below: this doesn't actually cancel the query due to a thread-local confusion]

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -143,7 +143,7 @@ class Database:
                     result = e
             task.reply_queue.sync_q.put(result)
 
-    async def execute_fn(self, fn):
+    async def execute_fn(self, fn, time_limit=None):
         def in_thread():
             conn = getattr(connections, self.name, None)
             if not conn:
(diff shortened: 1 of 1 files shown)


simonw:
I can remove this code too: 
https://github.com/simonw/datasette/blob/6f41c8a2bef309a66588b2875c3e24d26adb4850/datasette/database.py#L190-L192

datasette/database.py
```diff
@@ -152,9 +152,17 @@ def in_thread():
...
+        executor = asyncio.get_event_loop().run_in_executor(self.ds.executor, in_thread)
+        try:
+            return await asyncio.wait_for(
+                executor,
+                timeout=(time_limit / 1000.0) if time_limit is not None else None,
+            )
+        except asyncio.TimeoutError:
+            conn = getattr(connections, self.name, None)
+            if conn:
+                conn.interrupt()
```
simonw, line 164: This will never work, because `connections` is a thread local: https://github.com/simonw/datasette/blob/fb2ad7ada0b86a7fe4a576fe23236757c41eb05e/datasette/database.py#L24

simonw:
This isn't actually working - the outer code attempts to send an `.interrupt()` call to the connection object via the `connections` thread-local, which doesn't work because it's a thread-local so the connection isn't visible to that code.

Need to figure out how to communicate with that thread properly.

Issue #502, opened 2019-06-06
Exporting sqlite database(s)?


chrismp:
I'm working on datasette from one computer. But if I want to work on it from another computer and want to copy the SQLite database(s) already on the Heroku datasette instance, how to I copy the database(s) to the second computer so that I can then update it and push to online via datasette's command line code that pushes code to Heroku?

simonw:
There should be a "Download SQLite DB" link at the bottom of each database page - e.g. on https://latest.datasette.io/fixtures 

<img width="815" alt="fixtures" src="https://user-images.githubusercontent.com/9599/59298300-1bdff080-8c3f-11e9-8d75-b3eb6e1ae385.png">

Does that solve this? Re-open the issue if it doesn't!

chrismp:
Hm, neither that text nor that link appear on the site

chrismp:
It appears that I cannot reopen this issue but the proposed solution did not solve it.

louispotok:
For what it's worth, if anyone finds this in the future, I was having the same issue. 

After digging through the code, it turned out that the database download is only available if it the db served in immutable mode, so `datasette serve -i xyz.db` rather than the doc's quickstart recommendation of `datasette serve xyz.db`.

PR #1291, opened 2021-04-03
Update docs: explain allow_download setting

louispotok:
This fixes one possible source of confusion seen in #502 and clarifies
when database downloads will be shown and allowed.


simonw:
This is great, thank you.

Issue #1305, opened 2021-04-22
Index view crashes when any database table is not accessible to actor


slygent:
Because of https://github.com/simonw/datasette/blob/main/datasette/views/index.py#L63, the ```tables``` dict built does not include invisible tables; however, if https://github.com/simonw/datasette/blob/main/datasette/views/index.py#L80 is reached (because table_counts was not successfully initialized, e.g. due to a very large database) then as db.get_all_foreign_keys() returns ALL tables, a KeyError will be raised.

This error can be recreated with the fixtures.db if any table is hidden, e.g. by adding something like                 ```"foreign_key_references": {
                    "allow": {}
                }``` to fixtures-metadata.json and deleting ```or not table_counts``` from https://github.com/simonw/datasette/blob/main/datasette/views/index.py#L77.

I'm not sure how to fix this error; perhaps by testing if the table is in the aforementions ```tables``` dict.

PR #1306, opened 2021-04-22
Possible fix for issue #1305

slygent:


simonw:
We need to add a test that illustrates the bug in #1305 so we can prove that this definitely fixes it.

simonw:
https://github.com/simonw/datasette/blob/115332ce76c0e867d9936406aaf4bcee6b1ef3cb/datasette/views/index.py#L77 helps here - looks like we can pass `?_sort=relationships` to trigger the bug.

simonw:
I have a test - I'll merge this and then add that test.

Issue #619, opened 2019-11-10
"Invalid SQL" page should let you edit the SQL


simonw:
https://latest.datasette.io/fixtures?sql=select%0D%0A++*%0D%0Afrom%0D%0A++%5Bfoo%5D

<img src="https://user-images.githubusercontent.com/9599/68550604-1ab31b00-03b9-11ea-8a33-5dc339b8c5f5.jpeg" width="500">

Would be useful if this page showed you the invalid SQL you entered so you can edit it and try again.

davidszotten:
just trying out datasette and quite like it, thanks! i found this issue annoying enough to have a go at a fix. have you any thoughts on a good approach? (i'm happy to dig in myself if you haven't thought about it yet, but wanted to check if you had an idea for how to fix when you raised the issue)

obra:
I've just run into this after crafting a complex query and discovered that hitting back loses my query.

Even showing me the whole bad query would be a huge improvement over the current status quo.

simonw:
Yeah that sucks. Bumping this up the priority list.

simonw:
This is a little tricky to solve, because of the location of the form and the need to return JSON as well as HTML. It would be weird if a JSON request came in and got back the standard output from https://latest.datasette.io/fixtures.json when they were expecting to get back JSON in the shape of https://latest.datasette.io/fixtures.json?sql=select%20*%20from%20sqlite_master

I'm going to return the HTML view that you would get for 0 results for a query - https://latest.datasette.io/fixtures?sql=select%201%20limit%200 - but with an error message.

simonw:
So the JSON (still served with a 500 code) will look something like this:
```json
{
  "database": "fixtures",
  "query_name": null,
  "rows": [],
  "truncated": false,
  "error": "Error message goes here",
  "columns": [],
  "query": {
    "sql": "the query that broke goes here",
    "params": {}
  },
  "private": false,
  "allow_execute_sql": true,
  "query_ms": 0.8716583251953125,
  "source": "tests/fixtures.py",
  "source_url": "https://github.com/simonw/datasette/blob/master/tests/fixtures.py",
  "license": "Apache License 2.0",
  "license_url": "https://github.com/simonw/datasette/blob/master/LICENSE"
}
```

simonw:
I'll add this to the succesful JSON format:
```json
{
  "ok": true,
  "error": null
}
```

simonw:
[...]

PR #1346, opened 2021-05-28
Re-display user's query with an error message if an error occurs

simonw:
Refs #619

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -646,7 +646,7 @@ class Datasette:
                 "is_memory": d.is_memory,
                 "hash": d.hash,
             }
-            for name, d in sorted(self.databases.items(), key=lambda p: p[1].name)
+            for name, d in self.databases.items()
             if name != "_internal"
         ]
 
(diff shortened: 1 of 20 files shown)


simonw:
The one test failure here actually illustrates a larger problem: if the user specifies `.json?_shape=array` but an error occurs, what should we do?

Prior to this change we return the following JSON:
```json
{
    "ok": false,
    "error": "You did not supply a value for binding 1.",
    "status": 500,
    "title": null
}
```
But this comes from the `handle_500` higher level code here: https://github.com/simonw/datasette/blob/eae3084b46e2c3931db12cdef79093ad0e644bce/datasette/app.py#L1251-L1263

simonw:
This is the current test failure, but it actually another problem that we don't have tests in place for errors with different formats and shapes:

```
    def test_magic_parameters_cannot_be_used_in_arbitrary_queries(magic_parameters_client):
        response = magic_parameters_client.get(
            "/data.json?sql=select+:_header_host&_shape=array"
        )
        assert 400 == response.status
>       assert "You did not supply a value for binding 1." == response.json["error"]
E       TypeError: list indices must be integers or slices, not str
```

PR #1368, opened 2021-06-07
DRAFT: A new plugin hook for dynamic metadata

brandonrobertz:
Note that this is a WORK IN PROGRESS!

This PR adds the following plugin hook:

    get_metadata(
      datasette=self, key=key, database=database, table=table,
      fallback=fallback
    )

This gets called when we're building our metdata for the rest of the system to use. Datasette merges whatever the plugins return with any local metadata (from metadata.yml/yaml/json) allowing for a live-editable dynamic Datasette. __A major design consideration is this: should Datasette perform the metadata merge? Or should Datasette allow plugins to perform any modifications themselves?__

As a security precation, local meta is *not* overwritable by plugin hooks. The workflow for transitioning to live-meta would be to load the plugin with the full metadata.yaml and save. Then remove the parts of the metadata that you want to be able to change from the file.

I have a WIP dynamic configuration plugin here, for reference: https://github.com/next-LI/datasette-live-config/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -380,6 +380,7 @@ class Datasette:
         return self.databases[name]
 
     def add_database(self, db, name=None):
+        new_databases = self.databases.copy()
         if name is None:
             # Pick a unique name for this database
             suggestion = db.suggest_name()
@@ -391,14 +392,18 @@ class Datasette:
             name = "{}_{}".format(suggestion, i)
             i += 1
         db.name = name
-        self.databases[name] = db
+        new_databases[name] = db
+        # don't mutate! that causes race conditions with live import
+        self.databases = new_databases
         return db
 
     def add_memory_database(self, memory_name):
         return self.add_database(Database(self, memory_name=memory_name))
 
     def remove_database(self, name):
-        self.databases.pop(name)
+        new_databases = self.databases.copy()
+        new_databases.pop(name)
+        self.databases = new_databases
 
     def setting(self, key):
         return self._settings.get(key, None)
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -418,9 +418,7 @@ class Datasette:
 
         for key, upd_value in updated.items():
             if isinstance(upd_value, dict) and isinstance(orig.get(key), dict):
-                orig[key] = self._metadata_recursive_update(
-                    orig[key], upd_value
-                )
+                orig[key] = self._metadata_recursive_update(orig[key], upd_value)
             else:
                 orig[key] = upd_value
         return orig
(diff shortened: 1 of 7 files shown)


brandonrobertz:
Note that if we went with a "update_metadata" hook, the hook signature would look something like this:

```
update_metadata(
  datasette=self, metadata=metadata, key=key, database=database, table=table,
  fallback=fallback
)
```

And the Datasette function `_metadata_recursive_update(self, orig, updated)` would disappear into the plugins.

I have a WIP dynamic configuration plugin here, for reference: https://github.com/next-LI/datasette-live-config/

simonw:
A few tests failed - Black, the test that checks the docs mention the new hook - th most interesting failing test looks like this one:
```
            updated_metadata["databases"]["fixtures"]["queries"]["magic_parameters"][
                "allow"
            ] = (allow if "query" in permissions else deny)
>           cascade_app_client.ds._metadata = updated_metadata
E           AttributeError: can't set attribute
```

datasette/app.py
```diff
@@ -415,7 +433,24 @@ def metadata(self, key=None, database=None, table=None, fallback=True):
         assert not (
             database is None and table is not None
         ), "Cannot call metadata() with table= specified but not database="
-        databases = self._metadata.get("databases") or {}
+        metadata = {}
+
+        # TODO: change this to update_metadata, pass meta back allow the
+        # plugins to mutate it?
```
simonw, line 439: Let's ditch the TODO comments before we merge the PR - they can live in PR comments the issue conversation instead.

datasette/app.py
```diff
@@ -424,7 +459,10 @@ def metadata(self, key=None, database=None, table=None, fallback=True):
                 table
             ) or {}
             search_list.insert(0, table_metadata)
-        search_list.append(self._metadata)
+
+        # TODO: get full data for here, we need to work down to tables from that
+        # as opposed to simply returning table data above?
```
simonw, line 464: I don't fully understand this comment

datasette/app.py
```diff
@@ -440,6 +478,10 @@ def metadata(self, key=None, database=None, table=None, fallback=True):
                 m.update(item)
             return m
 
+    @property
+    def _metadata(self):
+        return self.metadata()
+
```
simonw, line 484: If this is for backwards compatibility I don't think it's necessary - `_metadata` has never been part of Datasette's public documented API.

brandonrobertz:
If this is a concept ACK then I will move onto fixing the tests (adding new ones) and updating the documentation for the new plugin hook.

simonw:
Yes, let's move ahead with getting this into an alpha.

simonw:
The only test failure is Black. I'm going to merge this and then reformat.

PR #1447, opened 2021-08-25
Remove underscore from search mode parameter name

wragge:
The fulltext search documentation refers to the parameter as `searchmode` but the `metadata.json` example uses `search_mode`. The latter doesn't actually seem to work.


simonw:
Good catch, thanks!

Issue #1486, opened 2021-10-12
pipx installation instructions for plugins don't reference pipx inject


RhetTbull:
The datasette [installation instructions](https://github.com/simonw/datasette/blob/main/docs/installation.rst) discuss how to install with pipx, how to upgrade with pipx, and how to upgrade plugins with pipx but do not mention how to install a plugin with pipx.  You discussed this on your [blog](https://til.simonwillison.net/python/installing-upgrading-plugins-with-pipx) but looks like this didn't make it in when you updated the docs for pipx (#756). 

I'll submit a PR shortly to fix this.

PR #1487, opened 2021-10-12
Added instructions for installing plugins via pipx, #1486

RhetTbull:
Adds missing instructions for installing plugins via pipx


simonw:
Thanks for this!

Issue #878, opened 2020-06-30
BaseView should be a documented API for plugins to use


simonw:
Can be part of #870 - refactoring existing views to use `register_routes()`.

> I'm going to put the new `check_permissions()` method on `BaseView` as well. If I want that method to be available to plugins I can do so by turning that `BaseView` class into a documented API that plugins are encouraged to use themselves.
_Originally posted by @simonw in https://github.com/simonw/datasette/issues/832#issuecomment-651995453_

simonw:
The `BaseView` class does this for Datasette internals at the moment, but I'm not convinced it works as well as it could.

I'd like to turn this into a class that is documented and available to plugins as well.

simonw:
I think this is blocking #619

simonw:
This has been blocking things for too long.

If this becomes a documented pattern, things like adding a JSON output to https://github.com/dogsheep/dogsheep-beta becomes easier too.

simonw:
I came up with a slightly wild idea for this that would involve pytest-style dependency injection.

Prototype here: https://gist.github.com/simonw/496b24fdad44f6f8b7237fe394a0ced7

Copying from my private notes:

> Use the lazy evaluated DI mechanism to break up table view into different pieces eg for faceting
> 
> Use that to solve JSON vs HTML views
> 
> Oh here's an idea: what if the various components of the table view were each defined  as async functions.... and then executed using asyncio.gather in order to run the SQL queries in parallel? Then try benchmarking with different numbers of threads?
> 
> The async_call_with_arguments function could do this automatically for any awaitable dependencies
> 
> This would give me massively parallel dependency injection
> 
> (I could build an entire framework around this and call it c64)
> 
> Idea: arguments called eg "count" are executed and the result passed to the function. If called count_fn then a reference to the not-yet-called function is passed instead 
> 
> I'm not going to completely combine the views mechanism and the render hooks. Instead, the core view will define a bunch of functions used to compose the page and the render hook will have conditional access to those functions - which will otherwise be asyncio.gather executed directly by the HTML page version
> 
> Using asyncio.gather to execute facets and suggest facets in parallel would be VERY interesting 
> 
> suggest facets should be VERY cachable - doesn't matter if it's wrong unlike actual facets themselves
> 
> What if all Datasette views were defined in terms of dependency injection - and those dependency functions could themselves depend on others just like pytest fixtures. Everything would become composable and async stuff could execute in parallel
> 
[...]

PR #1512, opened 2021-11-16
New pattern for async view classes

simonw:
Refs #878 - starting out with the new `AsyncBase` class implementing a pytest-inspired `asyncio` parallel execution mechanism.

- [ ] Refactor `Table` view to use the new base class and run facets and suggest facets in parallel
- [ ] ...

### DIFF
--- a/datasette/utils/asyncdi.py
+++ b/datasette/utils/asyncdi.py
@@ -1,0 +1,101 @@
+import asyncio
+from functools import wraps
+import inspect
+
+try:
+    import graphlib
+except ImportError:
+    from . import vendored_graphlib as graphlib
+
+
+class AsyncMeta(type):
+    def __new__(cls, name, bases, attrs):
+        # Decorate any items that are 'async def' methods
+        _registry = {}
+        new_attrs = {"_registry": _registry}
+        for key, value in attrs.items():
+            if inspect.iscoroutinefunction(value) and not value.__name__ == "resolve":
+                new_attrs[key] = make_method(value)
+                _registry[key] = new_attrs[key]
+            else:
+                new_attrs[key] = value
+        # Gather graph for later dependency resolution
+        graph = {
+            key: {
+                p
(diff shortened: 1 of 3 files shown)


simonw:
I've decided to make the clever `asyncio` dependency injection opt-in - so you can either decorate with `@inject` or you can set `inject_all = True` on the class - for example:
```python
import asyncio
from datasette.utils.asyncdi import AsyncBase, inject


class Simple(AsyncBase):
    def __init__(self):
        self.log = []

    @inject
    async def two(self):
        self.log.append("two")

    @inject
    async def one(self, two):
        self.log.append("one")
        return self.log

    async def not_inject(self, one, two):
        return one + two


class Complex(AsyncBase):
    inject_all = True

    def __init__(self):
        self.log = []

    async def b(self):
        self.log.append("b")

    async def a(self, b):
        self.log.append("a")

    async def go(self, a):
        self.log.append("go")
        return self.log
```

simonw:
```
E     File "/home/runner/work/datasette/datasette/datasette/utils/vendored_graphlib.py", line 56
E       if (result := self._node2info.get(node)) is None:
E                  ^
E   SyntaxError: invalid syntax
```
Oh no - the vendored code I use has `:=` so doesn't work on Python 3.6! Will have to backport it more thoroughly.

simonw:
I made the changes locally and tested them with Python 3.6 like so:
```
cd /tmp
mkdir v
cd v
pipenv shell --python=python3.6
cd ~/Dropbox/Development/datasette
pip install -e '.[test]'
pytest tests/test_asyncdi.py
```

simonw:
I'm going to extract out the `asyncinject` stuff into a separate library.

simonw:
https://github.com/simonw/asyncinject version 0.1a0 is now live on PyPI: https://pypi.org/project/asyncinject/

simonw:
Closing this PR because I shipped the code in it as a separate library instead.

Issue #1518, opened 2021-11-19
Complete refactor of TableView and table.html template


simonw:
Split from #878. The current `TableView` class is by far the most complex part of Datasette, and the most difficult to work on: https://github.com/simonw/datasette/blob/0.59.2/datasette/views/table.py

In #878 I started exploring a new pattern for building views. In doing so it became clear that `TableView` is the first beast that I need to slay - if I can refactor that into something neat the pattern for building other views will emerge as a natural consequence.

I've been trying to build this as a `register_routes()` plugin, as originally suggested in #870 - though unfortunately it looks like those plugins can't replace existing Datasette default views at the moment, see #1517. [UPDATE: I was wrong about this, plugins can over-ride default views just fine]

I also know that I want to have a fully documented template context for `table.html` as a major step on the way to Datasette 1.0, see #1510.

All of this adds up to the `TableView` factor being a major project that will unblock a whole flurry of other things - so I'm going to work on that in this separate issue.

simonw:
Here's where I got to with my hacked-together initial plugin prototype - it managed to render the table page with some rows on it (and a bunch of missing functionality such as filters): https://gist.github.com/simonw/281eac9c73b062c3469607ad86470eb2

<img width="904" alt="fixtures__roadside_attractions__4_rows_and__11__Liked___Twitter" src="https://user-images.githubusercontent.com/9599/142557224-600f645c-149e-4095-ab2e-73be824f3623.png">

simonw:
Ideally I'd like to execute the existing test suite against the new implementation - that would require me to solve this so I can replace the view with the plugin version though:

- #1517

simonw:
I was wrong about that, you CAN over-ride default routes already.

simonw:
A (likely incomplete) list of features on the table page:

- [ ] Display table/database/instance metadata
- [ ] Show count of all results
- [ ] Display table of results
  - [ ] Special table display treatment for URLs, numbers
  - [ ] Allow plugins to modify table cells
- [ ] Show interface for filtering by columns and operations
- [ ] Show search box, support executing FTS searches
- [ ] Sort table by specified column
- [ ] Paginate table
- [ ] Show facet results
- [ ] Show suggested facets
- [ ] Display schema for table
  - [ ] Maybe it should show the SQL for the query too?

simonw:
[...]

Issue #473, opened 2019-05-19
Plugin hook: filters_from_request


simonw:
I meant to add this as part of the facets plugin mechanism but didn't quite get to it. This will allow plugins to register extra filters, as seen in `datasette/filters.py`:

https://github.com/simonw/datasette/blob/260085838887ee343f4d3b177c422e7aef5ade9d/datasette/filters.py#L83-L98

simonw:
This expands on the refactoring work from https://github.com/simonw/datasette/commit/6da567dda953c7ac0e5500f17d8e220467a3499e

simonw:
It would be nice if this plugin was passed the current database/table so it can decide to enable new filters only for specific tables. This will require a bit of refactoring because the filters list is static at the moment - it would instead have to be returned by a function that runs when the table view is rendered.

simonw:
I wonder if this is the right hook?

The more likely case is that we need a hook that registers a new type of lookup entirely - `?_spatial_within={geojson}` for example. These aren't necessarily attached to a specific column and may have entirely custom syntax - but they still need to be able to influence the query (like the `_where=` clause does) and the human-readable description of the page.

Is there a strong case for supporting both custom filter plugins AND custom table where plugins, or could those where plugins cover both bases?

simonw:
I'm having trouble coming up with interesting column-based filters which don't make sense to ship as default behaviour.

simonw:
Assuming I do go ahead with this plugin hook, the existing `InFilter` makes for a nice simple example that illustrates the two key methods: `.where_clause()` and `.human_clause()`:

```python
class InFilter(Filter):
    key = "in"
    display = "in"

    def split_value(self, value):
        if value.startswith("["):
            return json.loads(value)
        else:
            return [v.strip() for v in value.split(",")]

    def where_clause(self, table, column, value, param_counter):
        values = self.split_value(value)
        params = [":p{}".format(param_counter + i) for i in range(len(values))]
        sql = "{} in ({})".format(escape_sqlite(column), ", ".join(params))
        return sql, values

    def human_clause(self, column, value):
        return "{} in {}".format(column, json.dumps(self.split_value(value)))
```

simonw:
Here's an older version of what that custom table filtering plugin might look like: https://github.com/simonw/datasette/commit/5116c4ec8aed5091e1f75415424b80f613518dc6

I built this for the https://github.com/simonw/russian-ira-facebook-ads-datasette project.

[...]

PR #1559, opened 2021-12-16
filters_from_request plugin hook, now used in TableView

simonw:
New plugin hook, refs #473

Used it to extract the logic from TableView that handles _search and
_through and _where - refs #1518

### DIFF
--- a/datasette/filters.py
+++ b/datasette/filters.py
@@ -1,7 +1,172 @@
+from datasette import hookimpl
+from datasette.views.base import DatasetteError
+from datasette.utils.asgi import BadRequest
 import json
 import numbers
+from .utils import detect_json1, escape_sqlite, path_with_removed_args
 
-from .utils import detect_json1, escape_sqlite
+
+@hookimpl(specname="filters_from_request")
+def where_filters(request, database, datasette):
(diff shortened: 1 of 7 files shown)


simonw:
This already has tests and documentation, and I've used it to refactor out the logic for `?_where=` and `?_search=` and `?_through=`.

Do I like this enough to land it on `main`?

Also, I think I can still use it to refactor out the `Filters` code that implements `?col=x` and `?col__lt=5` and suchlike.

simonw:
Before I land this I'm going to build one prototype plugin against it to confirm that the new hook is useful in its current shape.

I'll add support for filtering a table by drawing on a map to https://datasette.io/plugins/datasette-leaflet-freedraw

simonw:
I've convinced myself that this plugin hook design is good through this `datasette-leaflet-freedraw` prototype: https://github.com/simonw/datasette-leaflet-freedraw/blob/e8a16a0fe90656b8d655c02881d23a2b9833281d/datasette_leaflet_freedraw/__init__.py

simonw:
This is now available to try out in Datasette 0.60a0: https://github.com/simonw/datasette/releases/tag/0.60a0

Issue #1561, opened 2021-12-17
add hash id to "_memory" url if hashed url mode is turned on and crossdb is also turned on


fgregg:
If hashed_url mode is turned on and crossdb is also turned on, then queries to _memory should have a hash_id. 

One way that it could work is to have the _memory hash be a hash of all the individual databases.

Otherwise, crossdb queries can get quit out of data if using aggressive caching.

simonw:
Oh that's an interesting solution, combining the hashes of all of the individual databases.

I'm actually not a big fan of `hashed_url` mode - I implemented it right at the start of the project because it felt like a clever hack, and then ended up making it not-the-default a few years ago:

- #418
- #419
- #421

I've since not found myself wanting to use it at all for any of my projects - which makes me nervous, because it means there's a pretty complex feature that I'm not using at all, so it's only really protected by the existing unit tests for it.

What I'd really like to do is figure out how to have hashed URL mode work entirely as a plugin - then I could extract it from Datasette core entirely (which would simplify a bunch of stuff) but people who find the optimization useful would be able to access it.

I'm not sure that the existing plugin hooks are robust enough to do that yet though.

fgregg:
interesting! i love this feature. this + full caching with cloudflare is really super!

simonw:
I ended up moving hashed URL mode out to a plugin in:
- #647

If you're still interested in using it with `_memory` please open an issue in that repo here: https://github.com/simonw/datasette-hashed-urls

PR #1582, opened 2021-12-28
don't set far expiry if hash is '000'

### LINKED ISSUE #1581: when hashed urls are turned on, the _memory db has improperly long-lived cache expiry
if hashed_urls are on, then a -000 suffix is added to the `_memory` database, and the cache settings are set just as if it was a normal hashed database.

in particular, this header is set:

`cache-control: max-age=31536000`

this is not appropriate because the `_memory-000` database isn't really hashed based on the contents of the databases (see #1561).

Either the cache-control header should be changed, or the _memory db should have a hash suffix that does depend on the contents of the databases.

fgregg:
This will close #1581.

I couldn't find any unit tests related to the testing hashed urls, and I know that you want to break that code out of the core application (#1561), so I'm not quite sure what you would like me to for testing.

### DIFF
--- a/datasette/views/base.py
+++ b/datasette/views/base.py
@@ -628,6 +628,9 @@ class DataView(BaseView):
 
         ttl = request.args.get("_ttl", None)
         if ttl is None or not ttl.isdigit():
+            # if a db does not have hash, then the "hash" defaults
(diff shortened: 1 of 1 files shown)


fgregg:
this has been obviated by the datasette-hashed-urls plugin

PR #1617, opened 2022-02-01
Ensure template_path always uses "/" to match jinja

### LINKED ISSUE #1545: Custom pages don't work on windows
It seems that custom pages don't work when put in templates/pages

To reproduce on datasette version 0.59.4  using PowerShell on WIndows 10 with Python 3.10.0

    mkdir -p templates/pages

    echo "hello world" >> templates/pages/about.html

Start datasette
   
    datasette --template-dir templates/

Navigate to [http://127.0.0.1:8001/about](url) and receive:
    
   Error 404:
   Database not found: about

cb160:
This PR shoudl fix #1545 

The existing code substituted / for \, assuming this was the right behaviour for windows. But on Windows, Jinja still uses / for the template list - See https://github.com/pallets/jinja/blob/896a62135bcc151f2997e028c5125bec2cb2431f/src/jinja2/loaders.py#L225

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1212,7 +1212,8 @@ class DatasetteRouter:
         else:
             # Is there a pages/* template matching this path?
             route_path = request.scope.get("route_path", request.scope["path"])
-            template_path = os.path.join("pages", *route_path.split("/")) + ".html"
+            # Ensure template_path always uses "/" to match jina
+            template_path = "pages" + route_path + ".html"
             try:
                 template = self.ds.jinja_env.select_template([template_path])
             except TemplateNotFound:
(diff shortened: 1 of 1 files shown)


simonw:
Aha I understand the problem now! https://github.com/pallets/jinja/issues/1378#issuecomment-812410922

> Jinja template names/paths are not always filesystem paths. So regardless of the OS Jinja always uses forward slashes.

simonw:
I was hoping to get the test suite running on Windows before merging this PR but that looks like it will be a BIG job, see:
- #1627

So I'm going to merge this one as-is for the moment.

Issue #1625, opened 2022-02-02
Try running tests against macOS and Windows in addition to Ubuntu


simonw:
I already do this for `sqlite-utils`: https://github.com/simonw/sqlite-utils/blob/3.22.1/.github/workflows/test.yml

Related:
- #1617
- #1545

PR #1626, opened 2022-02-02
Try test suite against macOS and Windows

simonw:
Refs #1625


simonw:
That broke on a macOS test: https://github.com/simonw/datasette/runs/5044036993?check_suite_focus=true

I'm going to remove macOS and just try Windows purely to see what happens there.

simonw:
OK, the tests do NOT pass against Windows! https://github.com/simonw/datasette/runs/5044105941

image

Issue #1439, opened 2021-08-17
Rethink how .ext formats (v.s. ?_format=) works before 1.0


simonw:
Datasette currently has surprising special behaviour for if a table name ends in `.csv` - which can happen when a tool like `csvs-to-sqlite` creates tables that match the filename that they were imported from.

https://latest.datasette.io/fixtures/table%2Fwith%2Fslashes.csv illustrates this behaviour: it links to `.csv` and `.json` that look like this:

- https://latest.datasette.io/fixtures/table%2Fwith%2Fslashes.csv?_format=json
- https://latest.datasette.io/fixtures/table%2Fwith%2Fslashes.csv?_format=csv&_size=max

Where normally Datasette would add the `.csv` or `.json` extension to the path component of the URL (as seen on other pages such as https://latest.datasette.io/fixtures/facet_cities) here the [path_with_format() function](https://github.com/simonw/datasette/blob/adb5b70de5cec3c3dd37184defe606a082c232cf/datasette/utils/__init__.py#L710) notices that there is already a `.` in the path and instead adds `?_format=csv` to the query string instead.

The problem with this mechanism is that it's pretty surprising. Anyone writing external code to Datasette who wants to get back the `.csv` or `.json` version giving the URL to a table page will need to know about and implement this behaviour themselves. That's likely to cause all kinds of bugs in the future.

simonw:
The challenge comes down to telling the difference between the following:

- `/db/table` - an HTML table page
- `/db/table.csv` - the CSV version of `/db/table`
- `/db/table.csv` - no this one is actually a database table called `table.csv`
- `/db/table.csv.csv` - the CSV version of `/db/table.csv`
- `/db/table.csv.csv.csv` and so on...

simonw:
An alternative solution would be to use some form of escaping for the characters that form the name of the table.

The obvious way to do this would be URL-encoding - but it doesn't hold for `.` characters. The hex for that is `%2E` but watch what happens with that in a URL:

```
# Against Cloud Run:
curl -s 'https://datasette.io/-/asgi-scope/foo/bar%2Fbaz%2E' | rg path
 'path': '/-/asgi-scope/foo/bar/baz.',
 'raw_path': b'/-/asgi-scope/foo/bar%2Fbaz.',
 'root_path': '',
# Against Vercel:
~ % curl -s 'https://til.simonwillison.net/-/asgi-scope/foo/bar%2Fbaz%2E' | rg path
[...]

PR #1648, opened 2022-03-05
Use dash encoding for table names and row primary keys in URLs

simonw:
Refs #1439.

### DIFF
--- a/datasette/templates/_table.html
+++ b/datasette/templates/_table.html
@@ -4,7 +4,7 @@
         <thead>
             <tr>
                 {% for column in display_columns %}
-                    <th {% if column.description %}data-column-description="{{ column.description }}" {% endif %}class="col-{{ column.name|to_css_class }}" scope="col" data-column="{{ column.name }}" data-column-type="{{ column.type }}" data-column-not-null="{{ column.notnull }}" data-is-pk="{% if column.is_pk %}1{% else %}0{% endif %}">
+                    <th {% if column.description %}data-column-description="{{ column.description }}" {% endif %}class="col-{{ column.name|to_css_class }}" scope="col" data-column="{{ column.name }}" data-column-type="{{ column.type.lower() }}" data-column-not-null="{{ column.notnull }}" data-is-pk="{% if column.is_pk %}1{% else %}0{% endif %}">
                         {% if not column.sortable %}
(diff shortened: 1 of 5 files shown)


simonw:
Change of plan: based on extensive conversations on Twitter - see https://github.com/simonw/datasette/issues/1439#issuecomment-1059851259 - I'm going to try a variant of this which is basically percent-encoding but with a hyphen instead of a percent symbol.

Reason being that the current scheme doesn't handle the case of `%` being part of the table name, which could cause weird breakage due to some proxies decoding percent encoding before it gets to Datasette.

simonw:
Updated documentation: 
image

simonw:
For consistency, I'm going to change how `?_next=` tokens work too. Right now they work like this:

https://github.com/simonw/datasette/blob/de810f49cc57a4f88e4a1553d26c579253ce4531/datasette/views/table.py#L501-L507

https://github.com/simonw/datasette/blob/de810f49cc57a4f88e4a1553d26c579253ce4531/datasette/utils/__init__.py#L114-L116

I'm going to change those to use dash-encoding instead.

I considered looking for `%` in those values and replacing that as `-` too, but since Datasette isn't 1.0 yet I'm going to risk breaking any pagination tokens that people might have saved away somewhere!

simonw:
Just spotted this: https://github.com/simonw/datasette/blob/de810f49cc57a4f88e4a1553d26c579253ce4531/datasette/views/base.py#L203-L216

Maybe the db name should use dash encoding too?

simonw:
>     * Maybe use dash encoding for database name too?

Yes, I'm going to do this. At the moment if a DB file is called `fixx%tures.db` when you run it in Datasette the path is `/fix%2525tures` - which is liable to break.

simonw:
I may have to do extra work here
```python
    def database(self, database, format=None):
        db = self.ds.databases[database]
        if self.ds.setting("hash_urls") and db.hash:
            path = self.path(
                f"{dash_encode(database)}-{db.hash[:HASH_LENGTH]}", format=format
            )
        else:
            path = self.path(dash_encode(database), format=format)
        return path
```
The URLs that incorporate a hash have a `dbname-hash` format - will that `-` in the middle there mess up the dash decoding mechanism? I think it will.

Issue #1661, opened 2022-03-15
Remove Hashed URL mode


simonw:
It's now handled by a plugin instead:
- #647
- https://github.com/simonw/datasette-hashed-urls/issues/3

https://github.com/simonw/datasette-hashed-urls

Sub-tasks:

- [x] Remove hashed URL mode implementation
- [x] Update documentation
- [x] Ensure `--setting hash_urls 1` shows a useful message

simonw:
This is going to simplify the code in the various view classes substantially:
- #1660

simonw:
If you attempt to use the [old setting](https://docs.datasette.io/en/stable/settings.html#hash-urls):

    datasette mydatabase.db --setting hash_urls 1

It should error with a message saying that the feature has been moved to a plugin.

simonw:
From https://docs.datasette.io/en/0.60.2/performance.html#hashed-url-mode

> You can enable these hashed URLs in two ways: using the [hash_urls](https://docs.datasette.io/en/0.60.2/settings.html#setting-hash-urls) configuration setting (which affects all requests to Datasette) or via the `?_hash=1` query string parameter (which only applies to the current request).

I'm going to drop` ?_hash=1` entirely.

simonw:
Here's the only bit of code that references that `_hash` mechanism:

https://github.com/simonw/datasette/blob/77a904fea14f743560af9cc668146339bdbbd0a9/datasette/views/base.py#L259-L265

And here's the test:

https://github.com/simonw/datasette/blob/77a904fea14f743560af9cc668146339bdbbd0a9/tests/test_api.py#L828-L854

simonw:
https://datasette.io/plugins/datasette-hashed-urls is released now.

simonw:
As part of this I'm going to get rid of this mechanism: https://github.com/simonw/datasette/blob/30e5f0e67c38054a8087a2a4eae3fc4d1779af90/datasette/views/base.py#L170-L173

Unwrapping `request.scope["url_route"]["kwargs"]` into keyword argument to view functions just made the code harder to follow.

simonw:
Surprisingly I managed to break https://latest.datasette.io/fixtures/custom_foreign_key_label while working on this change:

![CleanShot 2022-03-17 at 16 16 54@2x](https://user-images.githubusercontent.com/9599/158909271-717b65e8-cfcc-44c4-b1cc-f34478b0f803.png)

simonw:
[...]

PR #1664, opened 2022-03-17
Remove hashed URL mode

simonw:
Refs #1661.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1097,7 +1097,7 @@ class Datasette:
         )
         add_route(
             TableView.as_view(self),
-            r"/(?P<db_name>[^/]+)/(?P<table_and_format>[^/]+?$)",
+            r"/(?P<db_name>[^/]+)/(?P<table>[^\/\.]+)(\.[a-zA-Z0-9_]+)?$",
         )
         add_route(
             RowView.as_view(self),
--- a/datasette/views/base.py
+++ b/datasette/views/base.py
@@ -372,57 +372,33 @@ class DataView(BaseView):
 
         return AsgiStream(stream_fn, headers=headers, content_type=content_type)
 
-    async def get_format(self, request, database, args):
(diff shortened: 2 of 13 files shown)


simonw:
Surprisingly I managed to break https://latest.datasette.io/fixtures/custom_foreign_key_label while working on this change:

CleanShot 2022-03-17 at 16 16 54@2x

simonw:
My hunch is that this is broken because of this:

https://github.com/simonw/datasette/blob/30e5f0e67c38054a8087a2a4eae3fc4d1779af90/datasette/app.py#L1098-L1107

Note how the table uses `table_and_format` but the row uses just `table` - I think there's code that's getting confused by this.

simonw:
Probably caused by the convoluted code is `get_format()`: https://github.com/simonw/datasette/blob/30e5f0e67c38054a8087a2a4eae3fc4d1779af90/datasette/views/base.py#L466-L481

simonw:
Now almost everything is working except for foreign key expansion:

CleanShot 2022-03-18 at 16 41 39@2x

Using the debugger I tracked it down to this code:

https://github.com/simonw/datasette/blob/30e5f0e67c38054a8087a2a4eae3fc4d1779af90/datasette/views/table.py#L708-L715

Turns out `default_labels` there is `None` - and it's a parameter to that `data()` method: https://github.com/simonw/datasette/blob/30e5f0e67c38054a8087a2a4eae3fc4d1779af90/datasette/views/table.py#L325-L334

simonw:
Looks like I deleted this code: https://github.com/simonw/datasette/blob/77a904fea14f743560af9cc668146339bdbbd0a9/datasette/views/base.py#L490-L492

simonw:
Still need to remove it from the documentation and do something about that `hash_urls` setting.

simonw:
I'm going to land this and handle those in separate commits.

PR #1717, opened 2022-04-23
Add timeout option to Cloudrun build

wragge:
I've found that the Cloudrun build phase often hits a timeout limit with large databases. I believe the default timeout is 10 minutes. This pull request just adds a `--timeout` option to the cloudrun `publish` command and passes the value on to the build step.


simonw:
This is a neat feature, thanks!

PR #1757, opened 2022-06-11
feat: add a wildcard for _json columns

ytjohn:
This allows _json to accept a wildcard for when there are many JSON columns that the user wants to convert. I hope this is useful. I've tested it on our datasette and haven't ran into any issues. I imagine on a large set of results, there could be some performance issues, but it will probably be negligible for most use cases.

On a side note, I ran into an issue where I had to upgrade black on my system beyond the pinned version in setup.py. Here is the upstream issue <<https://github.com/psf/black/issues/2964> . I didn't include this in the PR yet since I didn't look into the issue too far, but I can if you would like.

### DIFF
--- a/datasette/renderer.py
+++ b/datasette/renderer.py
@@ -10,13 +10,13 @@ from datasette.utils.asgi import Response
 
 def convert_specific_columns_to_json(rows, columns, json_cols):
     json_cols = set(json_cols)
-    if not json_cols.intersection(columns):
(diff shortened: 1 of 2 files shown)


simonw:
Interesting feature idea.

It's a weird edge-case, but it's the kind of thing I worry about: what if the user has a column called `*`? This is technically allowed in SQLite, even though it's vastly unlikely.

On that basis, I think `?_json_all=1` would be a better design here than `?_json=*` - since it removes any possible ambiguity.

PR #1789, opened 2022-08-19
Add new entrypoint option to `--load-extension`

### LINKED ISSUE #1784: Include "entrypoint" option on `--load-extension`?
## Problem 

SQLite extensions have the option to define multiple "entrypoints" in each loadable extension. For example, the upcoming version of `sqlite-lines` will have 2 entrypoints: the default `sqlite3_lines_init` (which SQLite will automatically guess for) and `sqlite3_lines_noread_init`. The `sqlite3_lines_noread_init` version omits functions that read from the filesystem, which is necessary for security purposes when running untrusted SQL (which Datasette does).

(Similar multiple entrypoints will also be added for sqlite-http).

The `--load-extension` flag, however, doesn't give the option to specify a different entrypoint, so the default one is always used. 

## Proposal

I want there to be a new command line option of the `--load-extension` flag to specify a custom entrypoint like so:
```
datasette my.db \
  --load-extension ./lines0 sqlite3_lines0_noread_init
```

Then, under the hood, this line of code:

https://github.com/simonw/datasette/blob/7af67b54b7d9bca43e948510fc62f6db2b748fa8/datasette/app.py#L562

Would look something like this:

```python
 conn.execute("SELECT load_extension(?, ?)", [extension, entrypoint]) 
```

One potential problem: For backward compatibility, I'm not sure if Click allows cli flags to have variable number of options ("arity"). So I guess it could also use a `:` delimiter like `--static`:

```
datasette my.db \
  --load-extension ./lines0:sqlite3_lines0_noread_init
```

Or maybe even a new flag name?

```
datasette my.db \
  --load-extension-entrypoint ./lines0 sqlite3_lines0_noread_init
```

Personally I prefer the `:` option... and maybe even `--load-extension` -> `--load`? Definitely out of scope for this issue tho

```
datasette my.db \
  --load./lines0:sqlite3_lines0_noread_init
```

asg017:
Closes #1784 

The `--load-extension` flag can now accept an optional "entrypoint" value, to specify which entrypoint SQLite should load from the given extension. 

```bash
# would load default entrypoint like before
datasette data.db --load-extension ext

# loads the extensions with the "sqlite3_foo_init" entrpoint
datasette data.db --load-extension ext:sqlite3_foo_init

# loads the extensions with the "sqlite3_bar_init" entrpoint
datasette data.db --load-extension ext:sqlite3_bar_init
```

For testing, I added a small SQLite extension in C at `tests/ext.c`. If compiled, then pytest will run the unit tests in `test_load_extensions.py`to verify that Datasette loads in extensions correctly (and loads the correct entrypoints). Compiling the extension requires a C compiler, I compiled it on my Mac with:

```
gcc ext.c -I path/to/sqlite -fPIC -shared -o ext.dylib
```

Where `path/to/sqlite` is a directory that contains the SQLite amalgamation header files.

Re documentation: I added a bit to the help text for `--load-extension` (which I believe should auto-add to documentation?), and the existing extension documentation is spatialite specific. Let me know if a new extensions documentation page would be helpful!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,9 @@ ENV/
 .DS_Store
 node_modules
 .*.swp
+
+# In case someone compiled tests/ext.c for test_load_extensions, don't
+# include it in source control.
+tests/*.dylib
+tests/*.so
+tests/*.dll
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -559,7 +559,13 @@ class Datasette:
         if self.sqlite_extensions:
             conn.enable_load_extension(True)
             for extension in self.sqlite_extensions:
-                conn.execute("SELECT load_extension(?)", [extension])
+                # "extension" is either a string path to the extension 
+                # or a 2-item tuple that specifies which entrypoint to load. 
+                if isinstance(extension, tuple):
+                    path, entrypoint = extension
+                    conn.execute("SELECT load_extension(?, ?)", [path, entrypoint])
+                else:
+                    conn.execute("SELECT load_extension(?)", [extension])
         if self.setting("cache_size_kb"):
             conn.execute(f"PRAGMA cache_size=-{self.setting('cache_size_kb')}")
         # pylint: disable=no-member
--- a/datasette/cli.py
+++ b/datasette/cli.py
@@ -21,6 +21,7 @@ from .app import (
     pm,
 )
 from .utils import (
+    LoadExtension,
     StartupError,
     check_connection,
     find_spatialite,
@@ -128,9 +129,10 @@ def sqlite_extensions(fn):
     return click.option(
         "sqlite_extensions",
         "--load-extension",
+        type=LoadExtension(),
         envvar="SQLITE_EXTENSIONS",
         multiple=True,
-        help="Path to a SQLite extension to load",
+        help="Path to a SQLite extension to load, and optional entrypoint",
     )(fn)
 
 
--- a/datasette/utils/__init__.py
+++ b/datasette/utils/__init__.py
@@ -833,6 +833,17 @@ class StaticMount(click.ParamType):
             self.fail(f"{value} is not a valid directory path", param, ctx)
         return path, dirpath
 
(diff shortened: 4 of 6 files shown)


simonw:
This looks great!

You need to run `cog -r docs/*.rst` to update the docs, then add those to the PR.

simonw:
If you cherry-pick in the fix from #1778 it should fix the Read the Docs failure.

simonw:
If you wanted to be really ambitious about documentation a page that explained what it takes to actually run SQLite extensions in Python would be great. I'm still trying to figure out the best way to do that in macOS myself.

simonw:
I don't think we should block landing this PR on that though.

asg017:
Rebased, Read the docs failure should now now fixed

Re docs - ya that's a pretty ambitious page, I'm still not 100% sure what the best practices are/should be... Would be happy to make that page in a future PR

simonw:
Any tips on how I can get the GitHub Action `test.yml` workflow to compile this so that the test runs there?

I imagine I need an extra step before https://github.com/simonw/datasette/blob/663ac431fe7202c85967568d82b2034f92b9aa43/.github/workflows/test.yml#L31

Neatest thing would be to cache the compiled file between actions runs (in the Actions cache) so we don't have to add the SQLite build dependencies on most runs.

asg017:
@simonw to build the extension on ubuntu, you can run:

```
apt-get update && apt-get install libsqlite3-dev gcc
gcc ext.c -fPIC -shared -o ext.so
```

I'm not the best with Actions, but if you set the cache key to `ext.c`, run those two commands to download dependencies + compile to `ext.so`, then the unit test should pick it up and run it correctly. Let me know if you want me to update the PR with that added

simonw:
I'm going to merge  as-is and then figure out the GitHub Actions bit separately. Thanks!

simonw:
Looks like it's not actually necessary to `apt-get install` anything extra in order to compile that test extension:

image

PR #1794, opened 2022-08-27
fix word break in facets by adding ul.tight-bullets li word-break: break-all

dmr:
I noticed that long words break the layout of facets:

image

So I added CSS to add a line break. This is how the result looks now:

image

I don't know enough about facet edge cases to decide if this change might break other things but it looks better for me so maybe this is helpful.

----
:books: Documentation preview :books:: https://datasette--1794.org.readthedocs.build/en/1794/


simonw:
Thanks for this - I've seen this issue too. Good fix.

PR #1820, opened 2022-09-26
No limit csv

fgregg:
----
:books: Documentation preview :books:: https://datasette--1820.org.readthedocs.build/en/1820/


fgregg:
the pattern in this PR `max_returned_rows` control the maximum rows rendered through html and json, and the csv render bypasses that.

i think it would be better to have each of these different query renderers have more direct control for how many rows to fetch, instead of relying on the internals of the `execute` method.

generally, users will not want to paginate through tens of thousands of results, but often will want to download a full query as json or as csv.

Issue #1822, opened 2022-09-26
Switch to keyword-only arguments for a bunch of internal methods


simonw:
This is a good idea, and one that needs to happen before Datasette 1.0:

> While you are adding features, would you be future-proofing your APIs if you switched over some arguments over to keyword-only arguments or would that be too disruptive?
>
> Thinking out loud:
>
> ```
> async def render_template( 
>      self, templates, *, context=None, plugin_context=None, request=None, view_name=None 
> ): 
> ```
_Originally posted by @jefftriplett in https://github.com/simonw/datasette/issues/1817#issuecomment-1256781274_

simonw:
Everything on https://docs.datasette.io/en/stable/internals.html that uses keyword arguments should do this I think.

simonw:
A start:
```diff
diff --git a/datasette/utils/asgi.py b/datasette/utils/asgi.py
index 8a2fa060..41ade961 100644
--- a/datasette/utils/asgi.py
+++ b/datasette/utils/asgi.py
@@ -118,7 +118,7 @@ class Request:
         return dict(parse_qsl(body.decode("utf-8"), keep_blank_values=True))
 
     @classmethod
-    def fake(cls, path_with_query_string, method="GET", scheme="http", url_vars=None):
+    def fake(cls, path_with_query_string, *, method="GET", scheme="http", url_vars=None):
         """Useful for constructing Request objects for tests"""
         path, _, query_string = path_with_query_string.partition("?")
         scope = {
@@ -204,7 +204,7 @@ class AsgiWriter:
         )
 
 
-async def asgi_send_json(send, info, status=200, headers=None):
+async def asgi_send_json(send, info, *, status=200, headers=None):
     headers = headers or {}
     await asgi_send(
         send,
@@ -215,7 +215,7 @@ async def asgi_send_json(send, info, status=200, headers=None):
     )
 
 
-async def asgi_send_html(send, html, status=200, headers=None):
+async def asgi_send_html(send, html, *, status=200, headers=None):
     headers = headers or {}
     await asgi_send(
         send,
@@ -226,7 +226,7 @@ async def asgi_send_html(send, html, status=200, headers=None):
     )
 
 
-async def asgi_send_redirect(send, location, status=302):
+async def asgi_send_redirect(send, location, *, status=302):
     await asgi_send(
         send,
         "",
@@ -236,12 +236,12 @@ async def asgi_send_redirect(send, location, status=302):
     )
 
 
[...]

PR #1823, opened 2022-09-27
Keyword-only arguments for a bunch of internal methods

simonw:
Refs #1822

----
:books: Documentation preview :books:: https://datasette--1823.org.readthedocs.build/en/1823/


simonw:
I should update the documentation to reflect this change.

PR #1835, opened 2022-10-06
use inspect data for hash and file size

### LINKED ISSUE #1834: inspect data is not used for caching database hash
When databases are loaded,

https://github.com/simonw/datasette/blob/cb1e093fd361b758120aefc1a444df02462389a3/datasette/app.py#L257-L260

there is nothing preventing the rehashing of the database for immutable databases.

https://github.com/simonw/datasette/blob/cb1e093fd361b758120aefc1a444df02462389a3/datasette/database.py#L50-L53

what i might expect is that relevant values of `inspect_data` get passed to the `Database` class to prevent re-hashing?

With data that is many gigs large, this is a significant start up time.

fgregg:
`inspect_data` should already include the hash and the db file size, so this PR takes advantage of using those instead of always recalculating. should help a lot on startup with large DBs.

closes #1834

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -48,9 +48,13 @@ class Database:
         self._read_connection = None
         self._write_connection = None
         if not self.is_mutable and not self.is_memory:
-            p = Path(path)
-            self.hash = inspect_hash(p)
-            self.cached_size = p.stat().st_size
+            if self.ds.inspect_data and self.ds.inspect_data.get(self.name):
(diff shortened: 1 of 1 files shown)


simonw:
Wow, great catch! The whole point of inspect data was to avoid this kind of expensive operation on startup so this makes total sense - I had no idea Datasette was still trying to hash a giant file every time the server started.

simonw:
See also:
- https://github.com/simonw/datasette/pull/1837

PR #1837, opened 2022-10-06
Make hash and size a lazy property

fgregg:
Many apologies, @simonw. My previous PR #1835 did not really solve the problem because the name of the database is often not known to database object in the init method.

I took a cue from how you dealt with this issue and made hash a lazy property and did something similar with size.

----
:books: Documentation preview :books:: https://datasette--1837.org.readthedocs.build/en/1837/


simonw:
Yeah this is better, thanks!

PR #1838, opened 2022-10-07
Open Datasette link in new tab

ocdtrekkie:
This is technically a Sandstorm-specific fix (as external links do not work inside the grain frame), however, I think it is an improvement to the upstream project, so I wanted to propose it here rather than patching it in our package.

There's much opinions on the Internet about whether external links should open in a new tab by default or not, but I'd argue very few people who might click a "powered by" link intend to complete their interaction with the source page (a Datasette). And furthermore, users may be working within various queries or loading visualizations (navigating away when trying to plot a million GPS coordinates pretty much just resets your progress!), so linking away within the tab might be a frustrating or destructive act to one's work, even inadvertently.

original report: https://github.com/ocdtrekkie/datasette-sandstorm/issues/1

----
:books: Documentation preview :books:: https://datasette--1838.org.readthedocs.build/en/1838/

### DIFF
--- a/datasette/templates/_footer.html
+++ b/datasette/templates/_footer.html
@@ -1,4 +1,4 @@
-Powered by <a href="https://datasette.io/" title="Datasette v{{ datasette_version }}">Datasette</a>
+Powered by <a href="https://datasette.io/" target="_blank" title="Datasette v{{ datasette_version }}">Datasette</a>
 {% if query_ms %}&middot; Queries took {{ query_ms|round(3) }}ms{% endif %}
 {% if metadata %}
     {% if metadata.license or metadata.license_url %}&middot; Data license:


simonw:
The argument that has always convinced me NOT to use `target="_blank"` (even for links like this one) is that it breaks browser expectations.

If you click a link with `target="_blank" on it you get a new browser window... with a disabled back button. You have to then know to close that browser window in order to return to the previous page - as opposed to hitting the "back" button like usual.

You'll note that Datasette doesn't use `target="_blank"` even on URLs presented in database tables - like these ones: https://latest.datasette.io/fixtures/roadside_attractions

So I'm very firmly in the anti-target-blank camp!

This is the kind of change which I'd suggest implementing as a plugin. `datasette-external-links-new-windows` could run a bit of JavaScript on every page that looks for `<a href="...">` elements that link to off-domain pages and adds `target="_blank"` to them via the DOM.

That way people who like `target="_blank"` can have it!

ocdtrekkie:
Ooh, I didn't even think about links in tables! You're definitely right on the approach to this. It might also be a really good "stupidly simple" plugin for me to try to build myself, which could be fun.

simonw:
... and here's @ocdtrekkie's plugin! https://github.com/ocdtrekkie/datasette-external-links-new-tabs

Issue #1829, opened 2022-10-04
Table/database that is private due to inherited permissions does not show padlock


simonw:
I noticed that a table page that is private because the database or instance is private, e.g. this one:

<img width="982" alt="image" src="https://user-images.githubusercontent.com/9599/193947303-ed643dcd-a07c-4ffb-ab22-539c4c803206.png">

Is not displaying the padlock icon that indicates the table is not visible to the public.

<img width="969" alt="image" src="https://user-images.githubusercontent.com/9599/193947373-ddbdf398-ea67-4e21-b54d-60c48496701d.png">

Same issue for the database page too, which in this case is private due to `view-instance`.

simonw:
Here's the relevant code from the table page:

https://github.com/simonw/datasette/blob/4218c9cd742b79b1e3cb80878e42b7e39d16ded2/datasette/views/table.py#L215-L227

Note how `ensure_permissions()` there takes the table, database and instance into account... but the `private` assignment  (used to decide if the padlock should display or not) only considers the `view-table` check.

Here's the same code for the database page:

https://github.com/simonw/datasette/blob/4218c9cd742b79b1e3cb80878e42b7e39d16ded2/datasette/views/database.py#L139-L141

And for canned query pages:

https://github.com/simonw/datasette/blob/4218c9cd742b79b1e3cb80878e42b7e39d16ded2/datasette/views/database.py#L228-L240

simonw:
There's also a `check_visibility()` helper which I'm not using in these particular cases but which may be relevant. It's called like this:

https://github.com/simonw/datasette/blob/4218c9cd742b79b1e3cb80878e42b7e39d16ded2/datasette/views/database.py#L65-L77

And is defined here: https://github.com/simonw/datasette/blob/4218c9cd742b79b1e3cb80878e42b7e39d16ded2/datasette/app.py#L694-L710

simonw:
I think `check_visibility` should be changed to optionally accept the same list of tuples that can be passed to `ensure_permissions`.

simonw:
I'll finish this in a PR:
- https://github.com/simonw/datasette/pull/1842

simonw:
Useful test: if you sign in as root to https://latest.datasette.io/_internal/columns/_internal,columns,database_name you can see there's no padlock icon on that page or on https://latest.datasette.io/_internal/columns - fixing this bug should fix that.

simonw:
[...]

PR #1842, opened 2022-10-14
check_visibility can now take multiple permissions into account

simonw:
Refs #1829

- [x] Fix table page
- [x] Fix database page
- [x] Fix query page
- [x] Fix row page
- [x] Tests
- [x] Updated documentation for `check_visibility` method, to cover the new `permissions=` keyword argument

Also this fix is currently only applied on the table page - needs to be applied on database, row and query pages too.

----
:books: Documentation preview :books:: https://datasette--1842.org.readthedocs.build/en/1842/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -736,7 +736,7 @@ class Datasette:
         self,
         actor: dict,
         action: Optional[str] = None,
-        resource: Optional[str] = None,
+        resource: Optional[Union[str, Tuple[str, str]]] = None,
         permissions: Optional[
             Sequence[Union[Tuple[str, Union[str, Tuple[str, str]]], str]]
         ] = None,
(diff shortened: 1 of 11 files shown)


simonw:
I need to do one last round of manual testing before I merge this.

simonw:
I'm going to construct a `metadata.yml` which makes various databases and tables visible or invisible, then browse them using the root user.

`block-instance.yml`:

```yaml
allow:
  id: root
```
`block-database.yml`:
```yaml
databases:
  fixtures:
    allow:
      id: root
```
`block-table.yml`:
```yaml
databases:
  fixtures:
    tables:
      searchable:
        allow:
          id: root
```
`block-query.yml`:
```yaml
databases:
  fixtures:
    queries:
      two:
        sql: select 1 + 1
        allow:
          id: root
```
https://gist.github.com/simonw/2d007ebe43de46d44499c77a2a291756 - checkout that Gist to get all four.

I manually tested all four scenarios with root and non-root users and confirmed that they worked correctly and padlocks were shown in the right places.

Issue #1836, opened 2022-10-06
docker image is duplicating db files somehow


fgregg:
if you look into the docker image created by docker publish, the `datasette inspect` line is duplicating the db files.

here's the result of the inspect command:

<img width="490" alt="Screen Shot 2022-10-06 at 2 58 08 PM" src="https://user-images.githubusercontent.com/536941/194430909-9303ee1a-9bdd-4212-b54b-de28f43768d4.png">

fgregg:
i thought it was maybe to do with reading through all the files, but that does not seem to be the case

if i make a little test file like:

```python
# test_read.py
import hashlib
import sys
import pathlib

HASH_BLOCK_SIZE = 1024 * 1024

def inspect_hash(path):
    """Calculate the hash of a database, efficiently."""
    m = hashlib.sha256()
    with path.open("rb") as fp:
        while True:
            data = fp.read(HASH_BLOCK_SIZE)
            if not data:
                break
            m.update(data)

    return m.hexdigest()

inspect_hash(pathlib.Path(sys.argv[1]))
```

then a line in the Dockerfile like

```docker
RUN python test_read.py nlrb.db && echo "[]" > /etc/inspect.json
```

just produes a layer of `3B`

fgregg:
it's not that the inspect command is somehow changing the db files. if i set them to only read-only, the "inspect" layer still has the same very large size.

fgregg:
okay, some progress!! running some sql against a database file causes that file to get duplicated even if it doesn't apparently change the file.

make a little test script like this:

```python
import sqlite3
import sys

conn = sqlite3.connect(sys.argv[1])
cur = conn.cursor()
cur.execute('select count(*) from filing')
print(cur.fetchone())
```

then 

```docker
RUN python test_sql.py nlrb.db
```

produced a layer that's the same size as `nlrb.db`!!

fgregg:
fascinatingly, telling python to open sqlite in read only mode makes this layer have a size of 0

```python
# test_sql_ro.py
import sqlite3
import sys

db_name = sys.argv[1]
conn = sqlite3.connect(f'file:/app/{db_name}?mode=ro', uri=True)
cur = conn.cursor()
cur.execute('select count(*) from filing')
print(cur.fetchone())
```

that's quite weird because setting the file permissions to read only didn't do anything.

fgregg:
and if we try immutable mode, which is how things are opened by `datasette inspect` we duplicate the files!!!

```python
[...]

Issue #1480, opened 2021-10-04
Exceeding Cloud Run memory limits when deploying a 4.8G database


ghing:
When I try to deploy a 4.8G SQLite database to Google Cloud Run, I get this error message:

> Memory limit of 8192M exceeded with 8826M used. Consider increasing the memory limit, see https://cloud.google.com/run/docs/configuring/memory-limits

Unfortunately, the maximum amount of memory that can be allocated to an instance is 8192M.

Naively profiling the memory usage of running Datasette with this database locally on my MacBook shows the following memory usage (using Activity Monitor) when I just start up Datasette:

- Real Memory Size: 70.6 MB
- Virtual Memory Size: 4.51 GB
- Shared Memory Size: 2.5 MB
- Private Memory Size: 57.4 MB

I'm trying to understand if there's a query or other operation that gets run during container deployment that causes memory use to be so large and if this can be avoided somehow.

This is somewhat related to #1082, but on a different platform, so I decided to open a new issue.

ghing:
Thanks for the reply @simonw. What services have you had better success with than Cloud Run for larger database?

Also, what about my issue description makes you think there may be a workaround?

Is there any instrumentation I could add to see at which point in the deploy the memory usage spikes? Should I be able to see this whether it's running under Docker locally, or do you suspect this is Cloud Run-specific?

ghing:
I was looking through the Dockerfile-generation code to see if there was anything that would cause memory usage to be a lot during deployment. 

I noticed that the Dockerfile [runs `datasette --inspect`](https://github.com/simonw/datasette/blob/main/datasette/utils/__init__.py#L354). Is it possible that this is using a lot of memory usage?

Or would that come into play when running `gcloud builds submit`, not when it's actually deployed?

fgregg:
running into this as i continue to grow my labor data warehouse.

Here a CloudRun PM says the container size should **not** count against memory: https://stackoverflow.com/a/56570717

fgregg:
as a next step, i'll fetch the docker image from the google registry, and see what memory and disk usage looks like when i run it locally.

simonw:
I've been having success using Fly recently for a project which I thought would be too large for Cloud Run. I wrote about that here:

- https://simonwillison.net/2022/Sep/5/laion-aesthetics-weeknotes/

fgregg:
thanks @simonw, i'll spend a little more time trying to figure out why this isn't working on cloudrun, and then will flip over to fly if i can't.

[...]

PR #1870, opened 2022-10-27
don't use immutable=1, only mode=ro

fgregg:
Opening db files in immutable mode sometimes leads to the file being mutated, which causes duplication in the docker image layers: see #1836, #1480

That this happens in "immutable" mode is surprising, because the sqlite docs say that setting this should open the database as read only. 

https://www.sqlite.org/c3ref/open.html

> immutable: The immutable parameter is a boolean query parameter that indicates that the database file is stored on read-only media. When immutable is set, SQLite assumes that the database file cannot be changed, even by a process with higher privilege, and so the database is opened read-only and all locking and change detection is disabled. Caution: Setting the immutable property on a database file that does in fact change can result in incorrect query results and/or [SQLITE_CORRUPT](https://www.sqlite.org/rescode.html#corrupt) errors. See also: [SQLITE_IOCAP_IMMUTABLE](https://www.sqlite.org/c3ref/c_iocap_atomic.html).

Perhaps this is a bug in sqlite?

----
:books: Documentation preview :books:: https://datasette--1870.org.readthedocs.build/en/1870/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -603,7 +603,7 @@ class Datasette:
                     continue
                 sql = 'ATTACH DATABASE "file:{path}?{qs}" AS [{name}];'.format(
                     path=db.path,
-                    qs="mode=ro" if db.is_mutable else "immutable=1",
+                    qs="mode=ro",
                     name=db_name,
                 )
                 conn.execute(sql)
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -81,13 +81,9 @@ class Database:
             return conn
         if self.is_memory:
             return sqlite3.connect(":memory:", uri=True)
-        # mode=ro or immutable=1?
-        if self.is_mutable:
-            qs = "?mode=ro"
-            if self.ds.nolock:
-                qs += "&nolock=1"
-        else:
-            qs = "?immutable=1"
+        qs = "?mode=ro"
+        if self.ds.nolock:
+            qs += "&nolock=1"
         assert not (write and not self.is_mutable)
         if write:
             qs = ""


fgregg:
Relevant sqlite forum thread: https://www.sqlite.org/forum/forumpost/02f7bda329f41e30451472421cf9ce7f715b768ce3db02797db1768e47950d48

fgregg:
as far as i can tell, [this is where the "immutable" argument is used](https://github.com/sqlite/sqlite/blob/c97bb14fab566f6fa8d967c8fd1e90f3702d5b73/src/pager.c#L4926-L4931) in sqlite:

```c
      pPager->noLock = sqlite3_uri_boolean(pPager->zFilename, "nolock", 0);
      if( (iDc & SQLITE_IOCAP_IMMUTABLE)!=0
       || sqlite3_uri_boolean(pPager->zFilename, "immutable", 0) ){
          vfsFlags |= SQLITE_OPEN_READONLY;
          goto act_like_temp_file;
      }
```

so it does set the read only flag, but then has a goto.

simonw:
Just saw your comment here: https://github.com/simonw/datasette/issues/1836#issuecomment-1272357976

> when you are running from docker, you **always** will want to run as `mode=ro` because the same thing that is causing duplication in the inspect layer will cause duplication in the final container read/write layer when `datasette serve` runs.

I don't understand this. My mental model of how Docker works is that the image itself is created using `docker build`... but then when the image runs later on (`docker run`) the image itself isn't touched at all.

Are you saying that I can build a container, but then when I run it and it does `datasette serve -i data.db ...` it will somehow modify the image, or create a new modified filesystem layer in the runtime environment, as a result of running that `serve` command?

simonw:
Saw your comment here too: https://github.com/simonw/datasette/issues/1480#issuecomment-1271101072

> switching from `immutable=1` to `mode=ro` completely addressed this. see https://github.com/simonw/datasette/issues/1836#issuecomment-1271100651 for details.

So maybe we need a special case for containers that are intended to be run using Docker - the ones produced by `datasette package` and `datasette publish cloudrun`? Those are cases where the `-i` option should actually be opened in read-only mode, not immutable mode.

Maybe a `datasette serve --irw data.db` option for opening a file in immutable-but-actually-read-only mode? Bit ugly though.

I should run some benchmarks to figure out if `immutable` really does offer significant performance benefits.

fgregg:
> Are you saying that I can build a container, but then when I run it and it does `datasette serve -i data.db ...` it will somehow modify the image, or create a new modified filesystem layer in the runtime environment, as a result of running that `serve` command?

Somehow, `datasette serve -i data.db` will lead to the `data.db` being modified, which will trigger a [copy-on-write](https://docs.docker.com/storage/storagedriver/#the-copy-on-write-cow-strategy) of `data.db` into the read-write layer of the container.

I don't understand **how** that happens.

jdangerx:
Hello! Resurrecting this issue since we're running into something similar with data.catalyst.coop as our database files have ballooned up to several GB. Our Cloud Run revisions now require huge amounts of RAM to start up without receiving a SIGBUS.

I'd love to see this fix merged in. It sounds like we want to make the immutable/read-only mode decision more flexible before doing so, so that we can use `ro` in Docker and `immutable` outside. If that sounds right, I'm happy to take a crack at adding that as a command-line flag or something that gets set automatically based on the expected execution environment.

PR #1965, opened 2022-12-18
Detect server start/stop more reliably.

janl:
This is useful, especially in testing, since your test hosts might not reliabliy start the server within two seconds, so we do a definite check before progressing.

By the same token, after `kill $server_pid` wait for the pid to be gone from the process list.

Since now the script can end prematurely, I also added a cleanup function to make sure the temporary certs are removed in any case.

n.b. this could also be done with the use of `trap 'fn' ERR` but that felt like a bit too much magic for this short a script.

----
:books: Documentation preview :books:: https://datasette--1965.org.readthedocs.build/en/1965/

### DIFF
--- a/tests/test_datasette_https_server.sh
+++ b/tests/test_datasette_https_server.sh
@@ -4,6 +4,10 @@
 python -m trustme
 # This creates server.pem, server.key, client.pem
 
+cleanup () {
+    rm server.pem server.key client.pem
+}
+
 # Start the server in the background
 datasette --memory \
     --ssl-keyfile=server.key \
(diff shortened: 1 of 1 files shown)


simonw:
This is great, thank you!

simonw:
Will link to this from my TIL shortly.

tests/test_datasette_https_server.sh
```diff
@@ -13,21 +17,48 @@ datasette --memory \
...
 # Shut down the server
 kill $server_pid
-sleep 1
+waiting=0
+#         show all pids
+#         |       find just the $server_pid
+#         |       |                  don’t match on the previous grep
+#         |       |                  |            we don’t need the output
+#         |       |                  |            |
+until ( ! ps ax | grep $server_pid | grep -v grep > /dev/null ); do
```
aspiers, line 50: This can be simplified in a few different ways:

```sh
until ! ps -p $server_pid > /dev/null; do

until ! [ -e /proc/$server_pid ]; do

until ! kill -0 $server_pid 2>/dev/null; do
```

These have varying degrees of portability - don't have a Mac to hand, but I'm guessing they all work on all modern systems.  They all assume that `$server_pid` cannot be the empty string, but I think that's safe in this case.

PR #1967, opened 2022-12-19
Add favicon to documentation

choldgraf:
I've been browsing the datasette documentation and found it hard to quickly locate tabs with many of them open, because it does not ship a favicon. So this PR:

- Grabs the favicon `.png` from datasette itself[^1]
- Adds it to the `_static/` folder
- Sets `html_favicon` to load it in the docs

[^1]: I also learned that Chrome can fetch favicons as an internal service! See `chrome://favicon/https://datasette.io/tools/github-to-sqlite`.


simonw:
Thanks for this!

Datasette has a favicon in this file: https://github.com/simonw/datasette/blob/2b4a8aa0833068da25072b6247bf349949ca00ee/datasette/static/favicon.png - so I'll copy that over and use it here as well.

simonw:
My Firefox tab before:

image

And after:

image

Issue #1099, opened 2020-11-15
Support linking to compound foreign keys


simonw:
Reported as a bug in #1098 because they caused 500 errors - but it would be even better if Datasette could hyperlink to related rows via compound foreign keys.

simonw:
My fix in deb0be4ae56f191f121239b29e83dd53b62d6305 for #1098 was to have Datasette deliberately pretend that compound foreign keys don't exist: https://github.com/simonw/datasette/blob/deb0be4ae56f191f121239b29e83dd53b62d6305/datasette/utils/__init__.py#L470-L495

This workaround will need to be rethought to implement real support for them.

simonw:
https://latest.datasette.io/_internal/foreign_keys (use https://latest.datasette.io/login-as-root first) is now a compound foreign key table:
```sql
CREATE TABLE foreign_keys (
        "database_name" TEXT,
        "table_name" TEXT,
        "id" INTEGER,
        "seq" INTEGER,
        "table" TEXT,
        "from" TEXT,
        "to" TEXT,
        "on_update" TEXT,
        "on_delete" TEXT,
        "match" TEXT,
        PRIMARY KEY (database_name, table_name, id, seq),
        FOREIGN KEY (database_name) REFERENCES databases(database_name),
        FOREIGN KEY (database_name, table_name) REFERENCES tables(database_name, table_name)
    );
```
Currently the `database_name` column becomes a link (because it's a single foreign key) but the `table_name` one remains a non-link:

<img width="1079" alt="_internal__foreign_keys__24_rows" src="https://user-images.githubusercontent.com/9599/102932110-87ba3080-4454-11eb-9ad5-b70a65129588.png">

My original idea for compound foreign keys was to turn both of those columns into links, but that doesn't fit here because `database_name` is already part of a different foreign key.

simonw:
Also need to solve displaying these links in the opposite direction:

https://latest.datasette.io/_internal/tables/fixtures,facet_cities

<img width="617" alt="_internal__tables" src="https://user-images.githubusercontent.com/9599/102944742-84cd3900-4470-11eb-95a0-bfaf00b17e82.png">

That page should link to lists of records in columns, foreign_keys and indexes.

fgregg:
I went ahead and spiked something together, in #2003

fgregg:
> My original idea for compound foreign keys was to turn both of those columns into links, but that doesn't fit here because `database_name` is already part of a different foreign key.

[...]

PR #2003, opened 2023-01-24
Show referring tables and rows when the referring foreign key is compound

fgregg:
sqlite foreign keys can be compound, but that is not as well supported by datasette as single column foreign keys.

in particular, 

1. in a table view, there is not a link from the row to the referenced row if the foreign key is compound
2. in a row view, there is no listing of tables and rows that refer to the focal row if those referencing foreign keys are compound.

Both of these issues are discussed in #1099. 

This PR only fixes the second one, because it's not clear what the right UX is for the first issue.

Screenshot 2023-01-24 at 19-47-40 nlrb bargaining_unit

Some things that might not be desirable about this approach.

1. it changes the external API, by changing `column` => `columns` and `other_column` => `other_columns` (see inline comment for more discussion.
2. There are various places where the plural foreign keys have to be checked for length and discarded or transformed to singular.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -899,8 +899,10 @@ class Datasette:
             fk = [
                 foreign_key
                 for foreign_key in foreign_keys
-                if foreign_key["column"] == column
+                if foreign_key["columns"][0] == column and len(foreign_keys) == 1
             ][0]
+            fk["column"] = fk["columns"][0]
+            fk["other_column"] = fk["other_columns"][0]
         except IndexError:
             return {}
         label_column = await db.label_column_for_table(fk["other_table"])
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -899,7 +899,8 @@ class Datasette:
             fk = [
                 foreign_key
                 for foreign_key in foreign_keys
-                if foreign_key["columns"][0] == column and len(foreign_keys) == 1
+                if foreign_key["columns"][0] == column
+                and len(foreign_key["columns"]) == 1
             ][0]
             fk["column"] = fk["columns"][0]
             fk["other_column"] = fk["other_columns"][0]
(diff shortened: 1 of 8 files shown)


datasette/app.py
```diff
@@ -899,8 +899,11 @@ async def expand_foreign_keys(self, database, table, column, values):
             fk = [
                 foreign_key
                 for foreign_key in foreign_keys
-                if foreign_key["column"] == column
+                if foreign_key["columns"][0] == column
```
fgregg, line 902: because we don't have a strong idea of how do display compound foreign keys in a table view, we have to do this operation in a few places.

tests/test_api.py
```diff
@@ -96,8 +96,8 @@ async def test_database_page(ds_client):
                 "incoming": [
                     {
                         "other_table": "roadside_attraction_characteristics",
-                        "column": "pk",
-                        "other_column": "characteristic_id",
+                        "columns": ["pk"],
```
fgregg, line 99: this is how the API changes.

tests/test_api.py
```diff
@@ -720,38 +740,43 @@ async def test_row_foreign_key_tables(ds_client):
     assert response.json()["foreign_key_tables"] == [
         {
             "other_table": "foreign_key_references",
-            "column": "id",
-            "other_column": "foreign_key_with_blank_label",
+            "columns": ["id"],
+            "other_columns": ["foreign_key_with_blank_label"],
             "count": 0,
             "link": "/fixtures/foreign_key_references?foreign_key_with_blank_label=1",
+            "other_columns_reference": "foreign_key_with_blank_label",
```
fgregg, line 747: this is also another API change. this got introduced so we could have compound keys look good in the row template, but we could do that through a custom filter instead.

fgregg:
@simonw, let me know what you think about this approach!

fgregg:
see this related discussion to a change in API in sqlite-utils https://github.com/simonw/sqlite-utils/pull/203#issuecomment-753567932

PR #2034, opened 2023-03-07
remove an unused `app` var in cli.py

wenhoujx:
this var `app` isn't actually used? unless init it does some side-effect outside of the event loop, idon't think it's necessary. 

Feel free to ignore this PR if the deleted line actually does something.

----
:books: Documentation preview :books:: https://datasette--2034.org.readthedocs.build/en/2034/


simonw:
Good catch, thanks.

wenhoujx:
@simonw any idea why the test coverage check fails?

Issue #2049, opened 2023-03-30
Custom SQL queries should use new JSON ?_extra= format


simonw:
Related:
- #262

I've made the change to the table view, now I need the new format to work for arbitrary SQL queries too.

simonw:
As part of this I should be able to figure out which bits of the new code I wrote for the table view should actually be shared with the query view. That stuff is mostly going to be from this commit: https://github.com/simonw/datasette/commit/d97e82df3c8a3f2e97038d7080167be9bb74a68d

Here's the existing QueryView class I need to replace:

https://github.com/simonw/datasette/blob/4c1e277edbd783d06840d3f9b20bf00783478ce4/datasette/views/database.py#L215-L532

simonw:
I'd really like to refactor all of the extras functions into a `datasette/extras.py` module. The table ones currently rely a LOT on local variables in scope though, so I would need to rewrite those such that EVERY dependency they take is passed to `asyncinject` explicitly.

simonw:
The default representation here can be even smaller.

For rows it's this:

```json
{
  "ok": true,
  "next": "d,v",
  "rows": [...]
}
```
For SQL queries I'm considering this:
```json
{
  "ok": true,
  "rows": [...]
}
```
I considered adding `"sql"` and `"params"` too, but on further thought those would be entirely a waste of bytes the majority of the time. If a user wants those they can request them with an `?_extra=query` as seen here:

http://localhost:8001/content/releases.json?_size=0&_extra=query

```json
{
  "ok": true,
  "next": null,
  "query": {
    "sql": "select html_url, id, author, node_id, tag_name, target_commitish, name, draft, prerelease, created_at, published_at, body, repo, reactions, mentions_count from releases order by id limit 1",
    "params": {}
  },
  "rows": []
}
```

simonw:
Completed in:
- https://github.com/simonw/datasette/pull/2769

PR #2053, opened 2023-04-05
WIP new JSON for queries

simonw:
Refs:
- #2049

TODO:

- [x] Read queries JSON
- [ ] Implement error display with `"ok": false` and an errors key
- [ ] Read queries HTML
- [ ] Read queries other formats (plugins)
- [ ] Canned read queries (dispatched to from table)
- [ ] Write queries (a canned query thing)
- [ ] Implement different shapes, refactoring to share code with table
- [ ] Implement a sensible subset of extras, also refactoring to share code with table
- [ ] Get all tests passing

----
:books: Documentation preview :books:: https://datasette--2053.org.readthedocs.build/en/2053/


simonw:
Table errors page currently does this:
```json
{
  "ok": false,
  "error": "no such column: blah",
  "status": 400,
  "title": "Invalid SQL"
}
```

simonw:
Uncommitted experimental code:
```diff
diff --git a/datasette/views/database.py b/datasette/views/database.py
index 455ebd1f..85775433 100644
--- a/datasette/views/database.py
+++ b/datasette/views/database.py
@@ -909,12 +909,13 @@ async def query_view(
     elif format_ in datasette.renderers.keys():
         # Dispatch request to the correct output format renderer
         # (CSV is not handled here due to streaming)
+        print(data)
         result = call_with_supported_arguments(
             datasette.renderers[format_][0],
             datasette=datasette,
-            columns=columns,
-            rows=rows,
-            sql=sql,
+            columns=data["rows"][0].keys(),
+            rows=data["rows"],
+            sql='',
             query_name=None,
             database=db.name,
             table=None,
@@ -923,7 +924,7 @@ async def query_view(
             # These will be deprecated in Datasette 1.0:
             args=request.args,
             data={
-                "rows": rows,
+                "rows": data["rows"],
             },  # TODO what should this be?
         )
         result = await await_me_maybe(result)
diff --git a/docs/index.rst b/docs/index.rst
index 5a9cc7ed..254ed3da 100644
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -57,6 +57,7 @@ Contents
    settings
    introspection
    custom_templates
+   template_context
    plugins
    writing_plugins
    plugin_hooks
```

simonw:
Now that I have the new `View` subclass from #2078 I want to use it to simplify this code.

Challenge: there are several things to consider here:

- The `/db` page without `?sql=` displays a list of tables in that database
- With `?sql=` it shows the query results for that query (or an error)
- If it's a `/db/name-of-canned-query` it works a bit like the query page, but executes a canned query instead of the `?sql=` query
- POST `/db/name-of-canned-query` is support for writable canned queries

simonw:
I'm going to entirely split canned queries off from `?sql=` queries - they share a bunch of code right now which is just making everything much harder to follow.

I'll refactor their shared bits into functions that they both call.

simonw:
Or maybe...

- `BaseQueryView(View)` - knows how to render the results of a SQL query
- `QueryView(BaseQueryView)` - renders from `?sql=`
- `CannedQueryView(BaseQueryView)` - renders for a named canned query

And then later perhaps:

- `RowQueryView(BaseQueryView)` - renders the `select * from t where pk = ?`
- `TableQueryView(BaseQueryView)` - replaces the super complex existing `TableView`

simonw:
I should split out a `canned_query.html` template too, as something that extends the `query.html` template.

simonw:
I should have an extra called `extra_html_context` which bundles together all of the weird extra stuff needed by the HTML template, and is then passed as the root context when the template is rendered (with the other stuff from extras patched into it).

simonw:
I think the hardest part of getting this working is dealing with the different formats.

Idea: refactor `.html` as a format (since it's by far the most complex) and tweak the plugin hook a bit as part of that, then use what I learn from that to get the other formats working.

simonw:
Another point of confusion is how `/content` sometimes serves the database index page (with a list of tables) and sometimes solves the results of a query.

I could resolve this by turning the information on the index page into extras, which can optionally be requested any time a query is run but default to being shown if there is no query.

simonw:
Worth noting that the `register_output_renderer()` is actually pretty easy to extend, because it returns a dictionary which could have more keys (like the required set of extras) added to it:
```python
@hookimpl
def register_output_renderer(datasette):
    return {
        "extension": "test",
        "render": render_demo,
        "can_render": can_render_demo,  # Optional
    }
```

simonw:
I'm abandoning this branch in favour of a fresh attempt, described here:
- https://github.com/simonw/datasette/issues/2109

I'll copy bits and pieces of this branch across as-needed.

PR #2056, opened 2023-04-11
GitHub Action to lint Python code with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The ruff Action uses minimal steps to run in ~5 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

----
:books: Documentation preview :books:: https://datasette--2056.org.readthedocs.build/en/2056/

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -10,8 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out repo
-      uses: actions/checkout@v2
-    - uses: actions/cache@v2
+      uses: actions/checkout@v3
+    - uses: actions/cache@v3
       name: Configure npm caching
       with:
         path: ~/.npm
--- a/.github/workflows/push_docker_tag.yml
+++ b/.github/workflows/push_docker_tag.yml
@@ -13,7 +13,7 @@ jobs:
   deploy_docker:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Build and push to Docker Hub
       env:
         DOCKER_USER: ${{ secrets.DOCKER_USER }}
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,15 @@
+# https://beta.ruff.rs
+name: ruff
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: pip install --user ruff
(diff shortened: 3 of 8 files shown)


simonw:
https://github.com/simonw/datasette/actions/runs/4664796647/jobs/8300596121?pr=2056 it's pretty fast - that finished in 9s.

simonw:
Here are the failures: https://github.com/simonw/datasette/actions/runs/4684460653/jobs/8300630794?pr=2056

CleanShot 2023-04-12 at 18 16 45@2x

simonw:
Cool - and now https://github.com/simonw/datasette/pull/2056/files is showing me those inline annotations:

image

simonw:
OK, I'm sold - this is a really neat improvement.

One thing to change in the PR: right now it runs `pip install --user ruff` on every commit, which hits PyPI to install the package.

I prefer to avoid hitting PyPI every time, so I like to use the GitHub Actions cache. My usual pattern for that looks like this:

https://github.com/simonw/datasette/blob/5890a20c374fb0812d88c9b0ef26a838bfa06c76/.github/workflows/test-pyodide.yml#L16-L21

Then a separate command that runs `pip install ...` will benefit from that cache.

Are you OK to make that change?

cclauss:
Ruff (written in Rust, not Python) is a 23MB executable so the time to download and pip install it dwarfs its runtime.

Let's run ruff with and without GitHub Actions pip cache to see the relative performance.

There are two great problems in computer science:
1. Naming things
2. Cache invalidation
3. Off-by-one errors

For 2., https://github.com/actions/setup-python#caching-packages-dependencies is vital reading.
Only exactly pinned requirements can be cached.  Currently in `setup.py` the only pinned dependencies are:
1. Sphinx==6.1.3
2. furo==2023.3.27
3. black==23.3.0
4. blacken-docs==1.13.0  # but unpinned elsewhere in `setup.py`

This means that there will be very few cache hits in the current actions.  See the link below to print out cache hits.
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#cache-hit

> PyPI uses Fastly's CDN to quickly serve content to end-users, allowing us to minimize our hosting infrastructure and obscure possible downtime.
* https://pypi.org/sponsors

I would be shocked if Fastly does not have CDN nodes in the same datacenters where GitHub Actions run so GHA requests to download `ruff` probably never hit a PyPI server.

cclauss:
Status?

Issue #2143, opened 2023-08-18
De-tangling Metadata before Datasette 1.0


asg017:
Metadata in Datasette is a really powerful feature, but is a bit difficult to work with. It was initially a way to add "metadata" about your "data" in Datasette instances, like descriptions for databases/tables/columns, titles, source URLs, licenses, etc. But it later became the go-to spot for other Datasette features that have nothing to do with metadata, like permissions/plugins/canned queries. 

Specifically, I've found the following problems when working with Datasette metadata:

1. Metadata cannot be updated without re-starting the entire Datasette instance.
2. The `metadata.json`/`metadata.yaml` has become a kitchen sink of unrelated (imo) features like plugin config, authentication config, canned queries
3. The Python APIs for defining extra metadata are a bit awkward (the `datasette.metadata()` class, `get_metadata()` hook, etc.)

## Possible solutions

Here's a few ideas of Datasette core changes we can make to address these problems. 

### Re-vamp the Datasette Python metadata APIs

The Datasette object has a single `datasette.metadata()` method that's a bit difficult to work with. There's also no Python API for inserted new metadata, so plugins have to rely on the `get_metadata()` hook.

The `get_metadata()` hook can also be improved - it doesn't work with async functions yet, so you're quite limited to what you can do.

(I'm a bit fuzzy on what to actually do here, but I imagine it'll be very small breaking changes to a few Python methods)

### Add an optional `datasette_metadata` table

Datasette should detect and use metadata stored in a new special table called `datasette_metadata`. This would be a regular table that a user can edit on their own, and would serve as a "live updating" source of metadata, than can be changed while the Datasette instance is running.

Not too sure what the schema would look like, but I'd imagine:

```sql
CREATE TABLE datasette_metadata(
  level text,
  target any,
  key text,
  value any,
  primary key (level, target)
)
```

Every row in this table would map to a single metadata "entry".

- `level` would be one of "datasette", "database", "table", "column", which is the "level" the entry describes. For example, `level="table"` means it is metadata about a specific table, `level="database"` for a specific database, or `level="datasette"` for the entire Datasette instance.
- `target` would "point" to the specific object the entry metadata is about, and would depend on what `level` is specific. 
  - `level="database"`: `target` would be the string name of the database that the metadata entry is about. ex `"fixtures"`
  - `level="table"`: `target` would be a JSON array of two strings. The first element would be the database name, and the second would be the table name. ex `["fixtures", "students"]`
[...]

Issue #2093, opened 2023-06-29
Proposal: Combine settings, metadata, static, etc. into a single `datasette.yaml` File


asg017:
Very often I get tripped up when trying to configure my Datasette instances. For example: if I want to change the port my app listen too, do I do that with a CLI flag, a `--setting` flag, inside `metadata.json`, or an env var? If I want to up the time limit of SQL statements, is that under `metadata.json` or a setting? Where does my plugin configuration go?

Normally I need to look it up in Datasette docs, and I quickly find my answer, but the number of places where "config" goes it overwhelming.

- Flat CLI flags like `--port`, `--host`, `--cors`, etc.
- `--setting`, like `default_page_size`, `sql_time_limit_ms` etc
- Inside `metadata.json`, including plugin configuration

Typically my Datasette deploys are extremely long shell commands, with multiple `--setting` and other CLI flags.

## Proposal: Consolidate all "config" into `datasette.toml`

I propose that we add a new `datasette.toml` that combines "settings", "metadata", and other common CLI flags like `--port` and `--cors` into a single file. It would be similar to "Cargo.toml" in Rust projects, "package.json" in Node projects, and "pyproject.toml" in Python, etc.

A sample of what it could look like:

```toml
# "top level" configuration that are currently CLI flags on `datasette serve`
[config]
port = 8020
host = "0.0.0.0"
cors = true

# replaces multiple `--setting` flags
[settings]
base_url = "/app/datasette/"
default_allow_sql = true
sql_time_limit_ms = 3500

# replaces `metadata.json`.
# The contents of datasette-metadata.json could be defined in this file instead, but supporting separate files is nice (since those are easy to machine-generate)
[metadata]
include="./datasette-metadata.json"

# plugin-specific 
[plugins]
[plugins.datasette-auth-github]
client_id = {env = "DATASETTE_AUTH_GITHUB_CLIENT_ID"}
client_secret = {env = "GITHUB_CLIENT_SECRET"}

[plugins.datasette-cluster-map]

latitude_column = "lat"
longitude_column = "lon"
```

## Pros
- Instead of multiple files and CLI flags, everything could be in one tidy file
- Editing config in a separate file is easier than editing CLI flags, since you don't have to kill a process + edit a command every time
- New users will know "just edit my `datasette.toml` instead of needing to learn metadata + settings + CLI flags
[...]

PR #2149, opened 2023-08-22
Start a new `datasette.yaml` configuration file, with settings support

asg017:
refs #2093 #2143 

This is the first step to implementing the new `datasette.yaml`/`datasette.json` configuration file. 

- The old `--config` argument is now back, and is the path to a `datasette.yaml` file. Acts like the `--metadata` flag.
- The old `settings.json` behavior has been removed.
- The `"settings"` key inside `datasette.yaml` defines the same `--settings` flags
- Values passed in `--settings` will over-write values in `datasette.yaml`

Docs for the Config file is pretty light, not much to add until we add more config to the file.

----
:books: Documentation preview :books:: https://datasette--2149.org.readthedocs.build/en/2149/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -242,6 +242,7 @@ class Datasette:
         cache_headers=True,
         cors=False,
         inspect_data=None,
+        config=None,
         metadata=None,
         sqlite_extensions=None,
         template_dir=None,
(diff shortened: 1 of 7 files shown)


simonw:
This is a really great start - tests pass, implementation looks clean, the new stubbed documentation page makes sense. Let's land it in `main` and iterate on it further in future PRs.

Issue #2153, opened 2023-08-24
Datasette --get --actor option


simonw:
I experimented with a prototype of this here:
- https://github.com/simonw/datasette/issues/2102#issuecomment-1691037971_

Which lets me run requests as if they belonged to a specific actor like this:

```bash
datasette fixtures.db --get '/fixtures/facetable.json' --actor '{
    "_r": {
        "r": {
            "fixtures": {
                "facetable": [
                    "vt"
                ]
            }
        }
    },
    "a": "user"
}'
```

Really useful for testing actors an `_r` options. Is this worth adding as a feature?

simonw:
Building that `"_r"` array is the main reason this would be useful, but it's also fiddly to get right.

`datasette create-token` has a design for that already: https://docs.datasette.io/en/1.0a4/authentication.html#datasette-create-token
```
datasette create-token root \
    --secret mysecret \
    --all view-instance \
    --all view-table \
    --database docs view-query \
    --resource docs documents insert-row \
    --resource docs documents update-row \
```
Adding imitations of those options to `datasette serve` would add a LOT of extra options, but it would also make it really convenient to attempt a request with a specific set of restrictions. Not sure if that would be worth the extra `--help` output or not.

simonw:
Another option: implement this as a plugin, providing a new command like `datasette get ...`

Or implement `datasette client get ...` as core commands or a plugin - except that clashes with the `datasette client` command that https://github.com/simonw/dclient adds (which is a tool for hitting remote Datasette instances, not running simulated queries through a local one).

simonw:
Annoying that `datasette client ...` makes a great name both for a plugin that executes simulated queries against a local database (thanks to its similarity to the existing `datasette.client` Python API) but is also the ideal name for a command for running commands as a client of an external Datasette instance!

simonw:
`datasette serve` currently only has a `--get` - for this to be really useful it needs to grow `--post` and maybe other verbs too.

Which is a good argument for moving this functionality to `datasette client get ...` instead.

simonw:
Documentation:
- https://docs.datasette.io/en/latest/cli-reference.html#datasette-get

Issue #2102, opened 2023-07-14
API tokens with view-table but not view-database/view-instance cannot access the table


simonw:
> Spotted a problem while working on this: if you grant a token access to view table for a specific table but don't also grant view database and view instance permissions, that token is useless.
>
> This was a deliberate design decision in Datasette - it's documented on https://docs.datasette.io/en/1.0a2/authentication.html#access-permissions-in-metadata
>
>> If a user cannot access a specific database, they will not be able to access tables, views or queries within that database. If a user cannot access the instance they will not be able to access any of the databases, tables, views or queries.
>
> I'm now second-guessing if this was a good decision.

_Originally posted by @simonw in https://github.com/simonw/datasette-auth-tokens/issues/7#issuecomment-1636031702_

simonw:
I think I made this decision because I was thinking about default deny: obviously if a user has been denied access to a database. It doesn't make sense that they could access tables within it.

But now that I am spending more time with authentication tokens, which default to denying everything, except for the things that you have explicitly listed, this policy, no longer makes as much sense.

simonw:
Relevant code: 
https://github.com/simonw/datasette/blob/0f7192b6154edb576c41b55bd3f2a3f53e5f436a/datasette/app.py#L822-L855

simonw:
I tried some code spelunking and came across https://github.com/simonw/datasette/commit/d6e03b04302a0852e7133dc030eab50177c37be7 which says:

> - If you have table permission but not database permission you can now view the table page

Refs:
- #832

simonw:
This might only be an issue with the code that checks `_r` on actors.

simonw:
Here's that crucial comment:

>  If _r is defined then we use those to further restrict the actor.
>
>Crucially, we only use this to say NO (return False) - we never use it to return YES (True) because that might over-ride other restrictions placed on this actor

simonw:
Confirmed that this is an issue with regular Datasette signed tokens as well. I created one on https://latest.datasette.io/-/create-token with these details:
```json
{
    "_r": {
        "r": {
            "fixtures": {
                "sortable": [
                    "vt"
                ]
            }
        }
    },
    "a": "root",
    "d": 3600,
    "t": 1689614483
}
```
Run like this:
```
[...]

PR #2154, opened 2023-08-24
Cascade for restricted token view-table/view-database/view-instance operations

simonw:
Refs:
- #2102

Also includes a prototype implementation of `--actor option` which I'm using for testing this, from:
- #2153

----
:books: Documentation preview :books:: https://datasette--2154.org.readthedocs.build/en/2154/

### DIFF
--- a/datasette/default_permissions.py
+++ b/datasette/default_permissions.py
@@ -190,9 +190,22 @@ def permission_allowed_actor_restrictions(datasette, actor, action, resource):
     # Special case for view-instance: it's allowed if there are any view-database
     # or view-table permissions defined
     if action == "view-instance":
+        all_rules = _r.get("a") or []
+        if (
+            "view-database" in all_rules
+            or "vd" in all_rules
+            or "view-table" in all_rules
+            or "vt" in all_rules
+        ):
+            return None
         database_rules = _r.get("d") or {}
         for rules in database_rules.values():
-            if "vd" in rules or "view-database" in rules:
+            if (
+                "vd" in rules
+                or "view-database" in rules
+                or "vt" in rules
+                or "view-table" in rules
+            ):
                 return None
         # Now check resources
         resource_rules = _r.get("r") or {}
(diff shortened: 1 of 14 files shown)


simonw:
Notes on manual testing so far - it looks like this might be working!
- https://github.com/simonw/datasette/issues/2102#issuecomment-1691824713

datasette/default_permissions.py
```diff
@@ -187,6 +187,64 @@ def permission_allowed_actor_restrictions(datasette, actor, action, resource):
...
+    if action == "view-table":
+        # Can view table if they have view-table in database or instance too
+        database_name = resource[0]
+        all_rules = _r.get("a") or []
+        if "view-table" in all_rules or "vt" in all_rules:
+            return None
+        database_rules = (_r.get("d") or {}).get(database_name) or {}
+        if "vt" in database_rules or "view-table" in database_rules:
+            return None
+
```
simonw, line 247: This code is pretty complicated. Maybe it would be simple and better if the fact that a permission cascades like this was modeled on the permission classes here instead:

https://github.com/simonw/datasette/blob/2e2825869fc2655b5fcadc743f6f9dec7a49bc65/datasette/default_permissions.py#L7-L38

simonw: Permissions is currently a `namedtuple`: https://github.com/simonw/datasette/blob/2e2825869fc2655b5fcadc743f6f9dec7a49bc65/datasette/permissions.py#L3-L6

It's used by a few plugins too (post-1.0 alpha) via this plugin hook: https://docs.datasette.io/en/1.0a4/plugin_hooks.html#register-permissions-datasette

simonw: I could add a `implies_can_view=False` property to that - if it's true, then it means that having this permission implies that you can view the database (if the permission `takes_database`) and the instance.

simonw: Or... I could have a special subclass of `Permission` called `PermissionImpliesView` which is reserved for use by core, and isn't something that's documented for plugins to register their own.

simonw: I'm going to upgrade `Permission` from a named tuple to a `dataclass`.

simonw:
Should this have `implies_can_view=True` too? Probably: https://github.com/simonw/datasette/blob/d64a9896f743f87b673d58859a0ec16685594e79/datasette/default_permissions.py#L20-L22

simonw:
Reminder that I also need to confirm that `insert-row` works if you have it at the instance level, the database level or the resource level in `_r`.

simonw:
Code for this might be cleaner with a `Restriction()` class that takes a `"_r"` dictionary to the constructor and can then answer questions like `.any_resource_has_permission("view-table")`.

simonw:
The code to refactor is this: https://github.com/simonw/datasette/blob/d64a9896f743f87b673d58859a0ec16685594e79/datasette/default_permissions.py#L181-L280

I'm going to turn that into a more general `restrictions_allow_action` function.

Issue #2164, opened 2023-08-30
Research ability to only load a specific list of plugins


simonw:
I'm going to try and get this working through an environment variable, so that you can start Datasette and it will only load a subset of plugins including those that use the `register_commands()` hook.

Initial research on this:
- https://github.com/pytest-dev/pluggy/issues/422

PR #2165, opened 2023-08-30
DATASETTE_LOAD_PLUGINS environment variable for loading specific plugins

simonw:
- #2164

TODO:

- [x] Automated tests
- [ ] Documentation
- [x] Make sure `DATASETTE_LOAD_PLUGINS=''` works for loading zero plugins

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,3 +50,7 @@ jobs:
       run: |
         # This fails on syntax errors, or a diff was applied
         blacken-docs -l 60 docs/*.rst
+    - name: Test DATASETTE_LOAD_PLUGINS
+      run: |
+        pip install datasette-init datasette-json-html
+        tests/test-datasette-load-plugins.sh
--- a/datasette/plugins.py
+++ b/datasette/plugins.py
@@ -23,15 +23,15 @@ DEFAULT_PLUGINS = (
 pm = pluggy.PluginManager("datasette")
 pm.add_hookspecs(hookspecs)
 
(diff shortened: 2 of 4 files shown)


simonw:
Testing this is going to be a bit of a pain.

I think I'll add a whole separate test block to CI which installs a couple of plugins and then exercises this feature using `datasette plugins`.

I'll use `datasette-init` and `datasette-json-html` just because they are small and simple.

simonw:
Slight weirdness: I noticed that the output from the `datasette plugins` command looks like this for plugins loaded with the new environment variable:
```json
    {
        "name": "datasette_pretty_json",
        "static": false,
        "templates": false,
        "version": null,
        "hooks": [
            "render_cell"
        ]
    },
```
That should ideally be `datasette-pretty-json`, not `datasette_pretty_json`.

simonw:
The `load_setuptools_entrypoints()` function in Pluggy [does this](https://github.com/pytest-dev/pluggy/blob/0b41c9766508a46ae666cf281684df3164b3e2a9/src/pluggy/_manager.py#L376):

```python
            for ep in dist.entry_points:
                if (
                    ep.group != group
                    or (name is not None and ep.name != name)
                    # already registered
                    or self.get_plugin(ep.name)
                    or self.is_blocked(ep.name)
                ):
                    continue
                plugin = ep.load()
                self.register(plugin, name=ep.name)
```

simonw:
Documentation preview: https://github.com/simonw/datasette/blob/6321c9c055a640ed6ea98e231dc5813dcde1f773/docs/plugins.rst#controlling-which-plugins-are-loaded

PR #2174, opened 2023-09-06
Use $DATASETTE_INTERNAL in absence of --internal

asg017:
#refs 2157, specifically [this comment](https://github.com/simonw/datasette/issues/2157#issuecomment-1700291967)

Passing in `--internal my_internal.db` over and over again can get repetitive. 

This PR adds a new configurable env variable `DATASETTE_INTERNAL_DB_PATH`. If it's defined, then it takes place as the path to the internal database. Users can still overwrite this behavior by passing in their own `--internal internal.db` flag.

In draft mode for now, needs tests and documentation. 

Side note: Maybe we can have a sections in the docs that lists all the "configuration environment variables" that Datasette respects? I did a quick grep and found:

- `DATASETTE_LOAD_PLUGINS`
- `DATASETTE_SECRETS`

----
:books: Documentation preview :books:: https://datasette--2174.org.readthedocs.build/en/2174/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -310,9 +310,6 @@ class Datasette:
                 Database(self, file, is_mutable=file not in self.immutables)
             )
 
-        if internal is None and "DATASETTE_INTERNAL_DB_PATH" in os.environ:
-            internal = os.environ.get("DATASETTE_INTERNAL_DB_PATH")
-
         self.internal_db_created = False
         if internal is None:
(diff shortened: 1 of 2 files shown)


simonw:
We can use this here: https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables

simonw:
`DATASETTE_INTERNAL` would be more consistent with `DATASETTE_SECRET`.

Issue #538, opened 2019-07-03
Mechanism for secrets in plugin configuration


simonw:
See https://github.com/simonw/datasette-auth-github/issues/1

We need a mechanism where by plugins can tap into "secret" config options without exposing them in the visible metadata.json (where plugin configs currently live, see https://datasette.readthedocs.io/en/stable/plugins.html#plugin-configuration )

simonw:
Initial syntax suggestion:
```json
{
    "title": "datasette-auth-github demo",
    "plugins": {
        "datasette-auth-github": {
            "client_id": "986f5d837b45e32ee6dd",
            "client_secret": {"$env": "GITHUB_CLIENT_SECRET"}
        }
    }
}
```

simonw:
Another useful option is the ability to load secrets from a file. This allows the file to have permissions set on it to only be read by the Datasette user. It also interacts well with the Kubernetes secrets mechanism, which is file-based.
```json
{
    "plugins": {
        "datasette-auth-github": {
            "client_id": "986f5d837b45e32ee6dd",
            "client_secret": {"$file": "/secrets/github-client-secret"}
        }
    }
}

simonw:
Re-opening this because I messed it up: the secret options are still visible in `/-/metadata` because I mutate the dictionary in place!

https://github.com/simonw/datasette/blob/a2d45931935f6bb73605a94afedf9e78308c95d6/datasette/app.py#L273-L279

PR #2183, opened 2023-09-11
`datasette.yaml` plugin support

asg017:
Part of #2093

In #2149 , we ported over `"settings.json"` into the new `datasette.yaml` config file, with a top-level `"settings"` key. This PR ports over plugin configuration into top-level `"plugins"` key, as well as nested database/table plugin config.

From now on, no plugin-related configuration is allowed in `metadata.yaml`, and must be in `datasette.yaml` in this new format. This is a pretty significant breaking change. Thankfully, you should be able to copy-paste your legacy plugin key/values into the new `datasette.yaml` format.

An example of what `datasette.yaml` would look like with this new plugin config:

```yaml

plugins:
  datasette-my-plugin:
    config_key: value

databases:
  fixtures:
      plugins: 
        datasette-my-plugin:
          config_key: fixtures-db-value
    tables:
      students:
        plugins:
          datasette-my-plugin:
            config_key: fixtures-students-table-value

```

As an additional benefit, this now works with the new `-s` flag:

```bash
datasette --memory -s 'plugins.datasette-my-plugin.config_key' new_value
```

Marked as a "Draft" right now until I add better documentation. We also should have a plan for the next alpha release to document and publicize this change, especially for plugin authors (since their docs will have to change to say `datasette.yaml` instead of `metadata.yaml`

----
:books: Documentation preview :books:: https://datasette--2183.org.readthedocs.build/en/2183/

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -114,6 +114,7 @@ def make_app_client(
     inspect_data=None,
     static_mounts=None,
     template_dir=None,
+    config=None,
     metadata=None,
     crossdb=False,
 ):
@@ -158,6 +159,7 @@ def make_app_client(
             memory=memory,
             cors=cors,
             metadata=metadata or METADATA,
+            config=config or CONFIG,
             plugins_dir=PLUGINS_DIR,
             settings=settings,
             inspect_data=inspect_data,
@@ -294,7 +296,14 @@ def generate_sortable_rows(num):
             "sortable_with_nulls_2": rand.choice([None, rand.random(), rand.random()]),
             "text": rand.choice(["$null", "$blah"]),
         }
-
+CONFIG = {
+  "plugins": {
+      "name-of-plugin": {"depth": "root"},
+      "env-plugin": {"foo": {"$env": "FOO_ENV"}},
+      "env-plugin-list": [{"in_a_list": {"$env": "FOO_ENV"}}],
+      "file-plugin": {"foo": {"$file": TEMP_PLUGIN_SECRET_FILE}},
+  },
+}
 
 METADATA = {
     "title": "Datasette Fixtures",
@@ -306,12 +315,6 @@ METADATA = {
     "about": "About Datasette",
     "about_url": "https://github.com/simonw/datasette",
     "extra_css_urls": ["/static/extra-css-urls.css"],
-    "plugins": {
-        "name-of-plugin": {"depth": "root"},
-        "env-plugin": {"foo": {"$env": "FOO_ENV"}},
-        "env-plugin-list": [{"in_a_list": {"$env": "FOO_ENV"}}],
-        "file-plugin": {"foo": {"$file": TEMP_PLUGIN_SECRET_FILE}},
-    },
     "databases": {
         "fixtures": {
             "description": "Test tables description",
@@ -768,6 +771,7 @@ def assert_permissions_checked(datasette, actions):
     type=click.Path(file_okay=True, dir_okay=False),
 )
 @click.argument("metadata", required=False)
+@click.argument("config", required=False)
 @click.argument(
     "plugins_path", type=click.Path(file_okay=False, dir_okay=True), required=False
 )
@@ -782,7 +786,7 @@ def assert_permissions_checked(datasette, actions):
     type=click.Path(file_okay=True, dir_okay=False),
     help="Write out second test DB to this file",
 )
-def cli(db_filename, metadata, plugins_path, recreate, extra_db_filename):
+def cli(db_filename, config, metadata, plugins_path, recreate, extra_db_filename):
     """Write out the fixtures database used by Datasette's test suite"""
     if metadata and not metadata.endswith(".json"):
         raise click.ClickException("Metadata should end with .json")
@@ -805,6 +809,10 @@ def cli(db_filename, metadata, plugins_path, recreate, extra_db_filename):
         with open(metadata, "w") as fp:
             fp.write(json.dumps(METADATA, indent=4))
         print(f"- metadata written to {metadata}")
+    if config:
+        with open(config, "w") as fp:
+            fp.write(json.dumps(CONFIG, indent=4))
+        print(f"- config written to {config}")
     if plugins_path:
         path = pathlib.Path(plugins_path)
         if not path.exists():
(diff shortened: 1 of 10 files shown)


simonw:
On thinking about this further, I'm fine releasing it as another alpha provided it causes Datasette to error loudly with an explanatory message if you attempt to load `-m metadata.json` at a metadata file that includes `"plugins"` configuration.

Something like this:
```bash
datasette -m metadata.json
```
Outputs:
```
Datasette no longer accepts plugin configuration in --metadata. Move your `"plugins"` configuration blocks to a separate file - we suggest `datasette.yaml` - and start Datasette with `datasette -c datasette.yaml`.
```
For added points, let's have it suggest `datasette.json` if they used `metadata.json` and `datasette.yaml` if they tried to use `metadata.yaml`.

asg017:
@simonw docs are finished, this is ready for review!

One thing: I added "Configuration" as a top-level item in the documentation site, at the very bottom. Not sure if this is the best, maybe it can be named "datasette.yaml Configuration" or something similar?

Mostly because "Configuration" by itself can mean many things, but adding "datasette.yaml" would make it pretty clear it's about that specific file, and is easier to scan. I'd also be fine with using "datasette.yaml" instead of "datasette.json", since writing in YAML is much more forgiving (and advanced users will know JSON is also supported)

Also, maybe this is a chance to consolidate the docs a bit? I think "Settings", "Configuration", "Metadata", and "Authentication and permissions" should possibly be under the same section. Maybe even consolidate the different Plugin pages that exist?

tests/fixtures.py
```diff
@@ -782,7 +798,7 @@ def assert_permissions_checked(datasette, actions):
     type=click.Path(file_okay=True, dir_okay=False),
     help="Write out second test DB to this file",
 )
-def cli(db_filename, metadata, plugins_path, recreate, extra_db_filename):
```
simonw, line 785: I like that you updated this too. Trying that myself:
```bash
python tests/fixtures.py --help
```
```
Usage: fixtures.py [OPTIONS] [DB_FILENAME] [METADATA] [CONFIG] [PLUGINS_PATH]

  Write out the fixtures database used by Datasette's test suite

Options:
  --recreate                Delete and recreate database if it exists
  --extra-db-filename FILE  Write out second test DB to this file
  --help                    Show this message and exit.
```
```bash
python tests/fixtures.py /tmp/demo/fixtures.db /tmp/demo/metadata.json /tmp/demo/config.json /tmp/demo/plugins
```
```
Test tables written to /tmp/demo/fixtures.db
- metadata written to /tmp/demo/metadata.json
- config written to /tmp/demo/config.json
  Wrote plugin: /tmp/demo/plugins/register_output_renderer.py
  Wrote plugin: /tmp/demo/plugins/view_name.py
  Wrote plugin: /tmp/demo/plugins/my_plugin.py
  Wrote plugin: /tmp/demo/plugins/messages_output_renderer.py
  Wrote plugin: /tmp/demo/plugins/sleep_sql_function.py
  Wrote plugin: /tmp/demo/plugins/my_plugin_2.py
```
```bash
find /tmp/demo
```
```
/tmp/demo
/tmp/demo/plugins
/tmp/demo/plugins/register_output_renderer.py
/tmp/demo/plugins/view_name.py
/tmp/demo/plugins/my_plugin.py
/tmp/demo/plugins/messages_output_renderer.py
/tmp/demo/plugins/sleep_sql_function.py
/tmp/demo/plugins/my_plugin_2.py
/tmp/demo/config.json
/tmp/demo/fixtures.db
/tmp/demo/metadata.json
```
```bash
cat /tmp/demo/config.json
```
```json
{
    "plugins": {
        "name-of-plugin": {
            "depth": "root"
        },
        "env-plugin": {
            "foo": {
                "$env": "FOO_ENV"
            }
        },
        "env-plugin-list": [
            {
                "in_a_list": {
                    "$env": "FOO_ENV"
                }
            }
        ],
        "file-plugin": {
            "foo": {
                "$file": "/var/folders/x6/31xf1vxj0nn9mxqq8z0mmcfw0000gn/T/plugin-secret"
            }
        }
    },
    "databases": {
        "fixtures": {
            "plugins": {
                "name-of-plugin": {
                    "depth": "database"
                }
            },
            "tables": {
                "simple_primary_key": {
                    "plugins": {
                        "name-of-plugin": {
                            "depth": "table",
                            "special": "this-is-simple_primary_key"
                        }
                    }
                },
                "sortable": {
                    "plugins": {
                        "name-of-plugin": {
                            "depth": "table"
                        }
                    }
                }
            }
        }
    }
}
```

tests/test_cli.py
```diff
@@ -238,6 +238,44 @@ def test_setting(args):
...
+        [
+            "--plugins-dir",
+            plugins_dir,
+            "--get",
+            "/_memory.json?sql=select+prepare_connection_args()",
+        ],
+    )
+    assert result.exit_code == 0, result.output
+    assert (
+        json.loads(result.output).get("rows")[0].get("prepare_connection_args()")
```
simonw, line 256: I couldn't remember how this worked, found the answer here:

https://github.com/simonw/datasette/blob/a4c96d01b27ce7cd06662a024da3547132a7c412/tests/plugins/my_plugin.py#L14-L27

tests/test_plugins.py
```diff
@@ -246,11 +243,6 @@ async def test_plugin_config_env_from_list(ds_client):
     assert [{"in_a_list": "FROM_ENVIRONMENT"}] == ds_client.ds.plugin_config(
         "env-plugin-list"
     )
-    # Ensure secrets aren't visible in /-/metadata.json
-    metadata = await ds_client.get("/-/metadata.json")
-    assert [{"in_a_list": {"$env": "FOO_ENV"}}] == metadata.json()["plugins"][
-        "env-plugin-list"
-    ]
```
simonw, line 253: This made me think. That `{"$env": "ENV_VAR"}` hack was introduced back here:

- https://github.com/simonw/datasette/issues/538

The problem it was solving was that metadata was visible to everyone with access to the instance at `/-/metadata` but plugins clearly needed a way to set secret settings.

Now that this stuff is moving to config, we have some decisions to make:

1. Add `/-/config` to let people see the configuration of their instance, and keep the `$env` trick for secret settings.
2. Say all configuration aside from metadata is secret and make `$env` optional or ditch it entirely.
3. Allow plugins to announce which of their configuration options are secret so we can automatically redact them from `/-/config`

I've found `/-/metadata` extraordinarily useful as a user of Datasette - it really helps me understand exactly what's going on if I run into any problems with a plugin, if I can quickly check what the settings look like.

So I'm leaning towards option 1 or 3.

simonw:
I'm going to land this and make any further documentation tweaks on `main`.

PR #2191, opened 2023-09-18
Move `permissions`,  `allow` blocks, canned queries and more out of `metadata.yaml` and into `datasette.yaml`

asg017:
The PR moves the following fields from `metadata.yaml` to `datasette.yaml`:

```
permissions
allow
allow_sql
queries
extra_css_urls
extra_js_urls
```

This is a significant breaking change that users will need to upgrade their `metadata.yaml` files for. But the format/locations are similar to the previous version, so it shouldn't be too difficult to upgrade.

One note: I'm still working on the Configuration docs, specifically the "reference" section. Though it's pretty small, the rest of read to review

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1317,7 +1317,7 @@ class Datasette:
         ):
             hook = await await_me_maybe(hook)
             collected.extend(hook)
-        collected.extend(self.metadata(key) or [])
+        collected.extend((self.config or {}).get(key) or [])
         output = []
         for url_or_dict in collected:
             if isinstance(url_or_dict, dict):
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -721,7 +721,9 @@ class Datasette:
         return self._app_css_hash
 
     async def get_canned_queries(self, database_name, actor):
-        queries = self.metadata("queries", database=database_name, fallback=False) or {}
+        queries = (
+            ((self.config or {}).get("databases") or {}).get(database_name) or {}
+        ).get("queries") or {}
         for more_queries in pm.hook.canned_queries(
             datasette=self,
             database=database_name,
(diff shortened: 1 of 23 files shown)


simonw:
I tested this locally for permissions like this. `datasette.yml`:

```yaml
databases:
  content:
    allow:
      id: root
```
Started Datasette like this:
```bash
datasette --root content.db pottery2.db -c datasette.yml
```
As root I could see this (note the padlock):

CleanShot 2023-09-20 at 15 17 02@2x

http://127.0.0.1:8001/-/metadata returned `{}` showing that the permissions must have come from the config file instead.

simonw:
This is one of the most interesting illustrative examples in the new code:

https://github.com/simonw/datasette/blob/f7bdedff779606466b580d8528e5a44509291002/tests/fixtures.py#L301-L349

Interesting to note that it now has canned queries in it, which include this bit:

https://github.com/simonw/datasette/blob/f7bdedff779606466b580d8528e5a44509291002/tests/fixtures.py#L341-L342

It looks like metadata, but in this case it's configuration. That blur between metadata and configuration at the canned query level still feels a little bit odd to me, but I still think we're going in the right direction with it.

Also interesting, from that same file:

https://github.com/simonw/datasette/blob/f7bdedff779606466b580d8528e5a44509291002/tests/fixtures.py#L351-L399

There are a few things in that metadata block that are arguably configuration, not metadata - for example:

https://github.com/simonw/datasette/blob/f7bdedff779606466b580d8528e5a44509291002/tests/fixtures.py#L360

I think `extra_css_urls` is definitely configuration, not metadata.

https://github.com/simonw/datasette/blob/f7bdedff779606466b580d8528e5a44509291002/tests/fixtures.py#L369-L395

Most of that stuff is arguably configuration too.

simonw:
The `{"units": {"distance": "m", "frequency": "Hz"}}` bit is for the units feature which I've half-disabled already and would like to remove before 1.0, though ideally turning that functionality into a plugin instead (if I can figure out how to do that).

Issue #2208, opened 2023-11-15
No suggested facets when a column named 'value' is included


rgieseke:
When a column named 'value' is included there are no suggested facets is shown as the query uses an alias of 'value'.

https://github.com/simonw/datasette/blob/452a587e236ef642cbc6ae345b58767ea8420cb5/datasette/facets.py#L168-L174

Currently the following is shown (from https://latest.datasette.io/fixtures/facetable)

![image](https://github.com/simonw/datasette/assets/198537/a919509a-ea88-461b-b25b-8b776720c7c5)

When I add a column named 'value' only the JSON facets are processed.

![image](https://github.com/simonw/datasette/assets/198537/092bd0b3-4c20-434e-88f8-47e2b8994a1d)

I think that not using aliases could be a solution (except if someone wants to use a column named `count(*)` though this seems to be unlikely). I'll open a PR with that.

There is also a TODO with a similar question in the same file. I have not looked into that yet.

https://github.com/simonw/datasette/blob/452a587e236ef642cbc6ae345b58767ea8420cb5/datasette/facets.py#L512

rgieseke:
Without aliases:

![image](https://github.com/simonw/datasette/assets/198537/d9703d3b-9733-4e87-9954-4fc60a07784a)

The proposed fix in #2209 also works when the 'value' column is actually facetable (just added another value in the 'value' column).

![image](https://github.com/simonw/datasette/assets/198537/a37a0a1a-c36a-4c78-bdce-01b582637cc6)

PR #2209, opened 2023-11-15
Fix query for suggested facets with column named value

rgieseke:
See discussion in https://github.com/simonw/datasette/issues/2208

----
:books: Documentation preview :books:: https://datasette--2209.org.readthedocs.build/en/2209/


rgieseke:
Whoops, looks like a forgot to check for other places where the 'facetable' table is used in the tests.

rgieseke:
Looks like tests are passing now but there is an issue with yaml loading and/or cog.

https://github.com/simonw/datasette/actions/runs/6879299298/job/18710911166?pr=2209

PR #2236, opened 2024-01-16
set link to download db to nofollow

fgregg:
This directs well-behaved web crawlers to not attempt to download the whole sqlite db, which is something you probably don't want a web crawler to do. 

----
📚 Documentation preview 📚: https://datasette--2236.org.readthedocs.build/en/2236/


simonw:
Thanks, this is a good idea.

PR #2238, opened 2024-01-30
Template slot family of plugin hooks - top_homepage() and others

simonw:
Refs:
- #1191

TODO:
- [x] `top_homepage`
- [x] `top_database`
- [x] `top_table`
- [x] `top_row`
- [x] `top_query`
- [x] `top_canned_query`

----
📚 Documentation preview 📚: https://datasette--2238.org.readthedocs.build/en/2238/

### DIFF
--- a/datasette/hookspecs.py
+++ b/datasette/hookspecs.py
@@ -181,10 +181,10 @@ def top_homepage(datasette, request):
 
 
 # @hookspec
-# def top_query(datasette, request, database, query):
+# def top_query(datasette, request, database, sql):
 #     """HTML to include at the top of the query results page"""
 
 
 # @hookspec
-# def top_canned_query(datasette, request, database, query):
+# def top_canned_query(datasette, request, database, query_name):
 #     """HTML to include at the top of the canned query page"""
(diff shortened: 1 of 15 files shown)


simonw:
Preview documentation: https://datasette--2238.org.readthedocs.build/en/2238/plugin_hooks.html#top-homepage-datasette-request

simonw:
Should these be wrapped in a `<div>` with a known ID or class I wonder? HTML currently looks like this:

https://github.com/simonw/datasette/blob/0c814ceb34841ec71f264cff1ca20d8ab767cddd/datasette/templates/index.html#L1-L12

simonw:
Also am I right putting these above the description/metadata section, or should they be below it?

I think above. Conceptually that information is strongly related to the table that is being displayed.

simonw:
I thought about maybe using a `{{ slot("top_homepage") }}` template function instead of the `{{ top_homepage() }}` thing that gets injected in to the context, but different slots take different arguments (e.g. the row one takes `top_row(datasette, request, database, table, row)`) so actually it's better to feed them in from the view function after all.

PR #2245, opened 2024-01-31
Run coverage on Python 3.12

simonw:
----
📚 Documentation preview 📚: https://datasette--2245.org.readthedocs.build/en/2245/


simonw:
That didn't seem to make a notable difference - in fact this took 4m47s to run the tests compared to 4m34s on an older Python 3.9 run.

I'm going to land it anyway, may as well do the upgrade.

Issue #2270, opened 2024-02-13
Use of md5 hash causes FIPS error, requires usedforsecurity=False


darugar:
Use of the md5 hashing algorithm is not permitted in FIPS systems, exhibiting the error:

```
else hashlib.md5(name.encode("utf8")).hexdigest()[:6],
ValueError: [digital envelope routines] unsupported
```

This can be fixed via `hashlib.md5(usedforsecurity=False)`:

in database.py:
```
return hashlib.md5(self.name.encode("utf8"), usedforsecurity=False).hexdigest()[:6]
```

in __init__.py:
```
    md5_suffix = hashlib.md5(s.encode("utf8"), usedforsecurity=False).hexdigest()[:6]
```

I can do a PR if that's easier - I've tested these updates locally and in FIPS environments and they work, but pytest does not pass tests (master is also failing the pytests in the same way).

Note that a similar issue exists in the `pint` library where `hashlib.blake2b` is used - in order to allow datasette to run in FIPS this needs to be modified to, for example, `hashlib.sha512` .

simonw:
There's one problem here: it looks like `usedforsecurity=False` was added in Python 3.9, but Datasette still supports Python 3.8 until at least its EOL in October 2024: https://devguide.python.org/versions/

I'm happy to add it with a Python version check or similar, but would the FIPS scanning system identify something like the following?

```python
import hashlib
import sys

def non_security_md5(text):
    try:
        return hashlib.md5(text.encode("utf8"), usedforsecurity=False).hexdigest()
    except TypeError:
        # usedforsecurity is not supported
        return hashlib.md5(text.encode("utf8")).hexdigest()
```

If that still trips the security filter I'm not sure what to do here - I want to be able to support Python 3.8.

simonw:
Since this is purely a cosmetic thing and we're pre-Datasette-1.0 I'd be OK swapping MD5 for SHA256 here to please the filter.

simonw:
Code in question:

https://github.com/simonw/datasette/blob/5d7997418664bcdfdba714c16bd5a67c241e8740/datasette/database.py#L73-L77

https://github.com/simonw/datasette/blob/5d7997418664bcdfdba714c16bd5a67c241e8740/datasette/utils/__init__.py#L705-L725

The CSS bit is actually a bigger problem, because changing that will change the CSS class name (and the name used for custom templates) for existing Datasette instances - which could result in customized templates or CSS breaking in ways that people might not easily notice.

[...]

PR #2271, opened 2024-02-14
Use MD5 usedforsecurity=False on Python 3.9 and higher

simonw:
Refs:
- #2270

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -1,7 +1,6 @@
 import asyncio
 from collections import namedtuple
 from pathlib import Path
-import hashlib
 import janus
 import queue
 import sys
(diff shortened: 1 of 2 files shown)


simonw:
I manually tested this fix in a FIPS Ubuntu Docker image using the steps outlined here and the tests passed too:
- https://github.com/simonw/datasette/issues/2270#issuecomment-1942981643

simonw:
Crucially the tests passed against Python 3.8, which demonstrates that this branch works: https://github.com/simonw/datasette/blob/2273a4bd4919f1ec89d2b5e818c42e0c199a9c08/datasette/utils/__init__.py#L1406-L1411

PR #2342, opened 2024-05-06
debugger: load 'ipdb' if present

myhro:
Transparently chooses between the IPython-enhanced `ipdb` or the standard `pdb`.

To me, personally, the tab-completion alone is already a killer-feature when debugging `locals()` variables.

What do you think of adding this option?

----
📚 Documentation preview 📚: https://datasette--2342.org.readthedocs.build/en/2342/


simonw:
I tested this like so:

```bash
datasette install ipdb
```
Then:
```bash
datasette --pdb --get /44
```
Output:
```
> /Users/simon/Dropbox/Development/datasette/datasette/app.py(1626)resolve_database()
   1625         except KeyError:
-> 1626             raise DatabaseNotFound(database_route)
   1627 

ipdb> c
<!DOCTYPE html>
<html lang="en">
```

simonw:
This is great, thanks very much!

Issue #2360, opened 2024-06-22
Proposal — Datasette JSON API changes for 1.0


asg017:
refs #1509

To prepare for Datasette 1.0, we want to standardize and document all the HTTP API endpoints that every Datasette instance offers. This proposal goes over single single standard JSON API endpoint we should implement + document for Datasette 1.0.

In general, there are very few breaking changes:

- The awkward `/.json`, (ie https://latest.datasette.io/.json) endpoint will be removed.
- The `/database/table.json?sql=select...` endpoint for getting JSON results of an SQL query would be replaced with `/database/-/query` , but we will redirect the former to the latter so it shouldn't be "breaking"
- the keys/values in a few seldom-used JSON endpoints will be updated

(I will be editing this top-level comment while the spec changes)

### `/.json` 

Will be deprecated.

### `/-/instance.json` 

Info about the entire Datasette instance.
	- Instance-level metadata (title, description, etc.)
	- A "initial look" at databases, tables, queries for the entire Datasette instance (first 100)

<details>
<summary>Show interface</summary>

```ts
interface Instance {

  // only provided when _extra=metadata
  metadata?: {
    title?: string;
    description?: string;
  }
  databases: {
    count: number;
    entries: InstanceDatabaseEntry[];
  }
  tables: {
    count: number;
    entries: InstanceTableEntry[];
  }
  views: {
    count: number;
    entries: InstanceViewEntry[];
  }
  queries: {
    count: number;
    entries: InstanceQueryEntry[];
  }
}

interface InstanceDatabaseEntry {
  database: string;
  name: string;
  // TODO
}
interface InstanceTableEntry {
  database: string;
  name: string;
  // TODO
}
interface InstanceViewEntry {
  database: string;
  name: string;
  // TODO
}
interface InstanceQueryEntry {
  database: string;
  name: string;
  // TODO
}
```

</details>

### `/-/databases.json`

A paginated view of all the databases in the instance

<details>
<summary>Show interface</summary>

```ts
interface InstanceDatabases {
  databases: [];
  // only when _extra=metadata
  metadata?: DatabaseMetadata;
  // URL params to next page, if any
  next?: string;
  // full URL to next page, if any
  next_url?: string;
}

interface InstanceDatabase {
  name: string;
}
```

</details>

### `/-/resources.json`

A paginated view of all the tables, views, and queries in each database for the entire instance.

<details>
<summary>Show interface</summary>

```ts
interface XXX {
  // URL params to next page, if any
[...]

Issue #1919, opened 2022-11-30
Intermittent test failure 


simonw:
https://github.com/simonw/datasette/actions/runs/3580503393/jobs/6022689591

```
        delete_response = await ds_write.client.post(
            "/data/{}/{}/-/delete".format(table, delete_path),
            headers={
                "Authorization": "***".format(write_token(ds_write)),
            },
        )
>       assert delete_response.status_code == 200
E       assert 404 == 200
E        +  where 404 = <Response [404 Not Found]>.status_code

/home/runner/work/datasette/datasette/tests/test_api_write.py:396: AssertionError
=========================== short test summary info ============================
FAILED tests/test_api_write.py::test_delete_row[compound_pk_table-row_for_create2-pks2-article,k] - assert 404 == 200
 +  where 404 = <Response [404 Not Found]>.status_code
```
This passes most of the time, but very occasionally fails - in this case in Python 3.7

It seems to only fail for the `article,k` compound primary key test.

simonw:
This is the test: https://github.com/simonw/datasette/blob/8404b21556d133c89eda4bd1bf5335ed9a0785d6/tests/test_api_write.py#L342-L401

I'm suspicious that there's a timing error of some sort but I can't think what it might be.

PR #2363, opened 2024-07-03
Introduce new `/$DB/-/query` endpoint, soft replaces `/$DB?sql=...`

asg017:
refs #2360 

This PR adds a new dedicated endpoint to querying a database in Datasette with SQL. Previously, one would use the `/dbname?sql=select...` convention to do this, but it shared the same endpoints as normal database. So this PR adds a dedicated `/dbname/-/query?sql=...` route specifically for this task. 

Old URLs that use the previous `/dbname?sql=...` endpoint will still work - we internally just redirect that to the new URL. 

----
📚 Documentation preview 📚: https://datasette--2363.org.readthedocs.build/en/2363/

### DIFF
--- a/docs/pages.rst
+++ b/docs/pages.rst
@@ -40,6 +40,12 @@ The JSON version of this page provides programmatic access to the underlying dat
 * `fivethirtyeight.datasettes.com/fivethirtyeight.json <https://fivethirtyeight.datasettes.com/fivethirtyeight.json>`_
 * `global-power-plants.datasettes.com/global-power-plants.json <https://global-power-plants.datasettes.com/global-power-plants.json>`_
 
+.. _QueryView:
+
+Queries
+========
+
+TODO
 .. _DatabaseView_hidden:
 
 Hidden tables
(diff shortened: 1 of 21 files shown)


simonw:
Spotted that the form on this page:

CleanShot 2024-07-15 at 09 57 06@2x

Still targets `/content` which then redirects. I'll push a fix for that.

docs/pages.rst
```diff
@@ -40,6 +40,12 @@ The JSON version of this page provides programmatic access to the underlying dat
 * `fivethirtyeight.datasettes.com/fivethirtyeight.json <https://fivethirtyeight.datasettes.com/fivethirtyeight.json>`_
 * `global-power-plants.datasettes.com/global-power-plants.json <https://global-power-plants.datasettes.com/global-power-plants.json>`_
 
+.. _QueryView:
+
+Queries
+========
+
+TODO
```
simonw, line 48: Let's fill this out before merginge.

simonw:
That test failure was the same intermittent failure we've seen on 3.8 in a few other places:
- #1919

PR #2386, opened 2024-08-07
skip over "queries" blocks when processing database-level metadata items

asg017:
Before metadata that looked like:

```yaml
databases:
  legislators:
    tables:
      offices:
        summary: office address or sumtin
    queries:
      millenntial_represetatives:
        summary: Social media accounts for current legislators

```

Would error with:

image

That's because when processing `databases.legislators` items, we skip over `tables` but not `queries`. We "skip" them because we processing them separately as "table-level" metadata items. 

Don't merge yet, want to test this more. Shouldn't we also have a block to process `queries` separately too? 

----
📚 Documentation preview 📚: https://datasette--2386.org.readthedocs.build/en/2386/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -454,7 +454,7 @@ class Datasette:
         # step 2: database-level metadata
         for dbname, db in self._metadata_local.get("databases", {}).items():
             for key, value in db.items():
-                if key == "tables":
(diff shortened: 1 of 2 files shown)


simonw:
We need a test for this. Do we need to do anything else before we land it?

PR #2391, opened 2024-08-14
don't hide virtual table, hide shadow tables.

### LINKED ISSUE #2296: Automatically hide type=shadow tables in SQLite 3.37 or higher
> For filtering virtual table's "shadow tables" (ex the FTS5 _content and most the spatialite tables),  you can use `pragma_table_list` (first appeared in SQLite 3.37 (2021-11-27), which has a `type` column that calls out `type="shadow"` tables https://www.sqlite.org/pragma.html#pragma_table_list

_Originally posted by @asg017 in https://github.com/simonw/datasette/issues/2104#issuecomment-1641082395_

asg017:
Closes #2296 

- Hide shadow tables by default
- Do NOT hide FTS5 virtual tables (but hide their shadow tables)

This will likely fail the SQLite 3.25 test

----
📚 Documentation preview 📚: https://datasette--2391.org.readthedocs.build/en/2391/

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -506,8 +506,8 @@ class Database:
                           OR substr(name, 1, 1) == '_'
                       ),
                       fts_suffixes AS (
-                        select column1 as suffix
-                        from (VALUES ('_data'), ('_idx'), ('_docsize'), ('_content'))
+                        SELECT column1 AS suffix
+                        FROM (VALUES ('_data'), ('_idx'), ('_docsize'), ('_content'), ('_config'))
                       ),
                       fts_names AS (
                         SELECT name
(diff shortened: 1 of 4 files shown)


simonw:
Let's mark the test that's going to fail in the older SQLite version as `@pytest.mark.skipif` - the failure just means that in old versions of SQLite there will be some visible tables that would have been hidden in newer versions, which is fine.

asg017:
@simonw ready for review!

It wasnt' as easy as skipping the test for SQLite 3.37 - since other tests break because of this change (ie table lists that sort by name + hidden status), I couldnt skip all those other non-related tests. So I added some logic that manually adds in FTS3 + FTS4 + FTS5 shadow tables by hard-coded their suffixes. The SQL is a bit nasty, happy to edit that if needed

simonw:
This looks great!

PR #2450, opened 2024-12-06
fix: support static files with spaces in their names

andrew-grechkin:
Browser encodes all non-ASCII characters in URL into percent encoding format.

Static files are served from the filesystem, and can have unsafe characters in their names.

To support such files the request `path` should be decoded.

----
📚 Documentation preview 📚: https://datasette--2450.org.readthedocs.build/en/2450/


brandonrobertz:
It would be really nice to get this merged. This is an *incredibly* annoying bug. Every time I have to do a fork and patch the static handler, or rename every single file and deal with the duplication problems it causes.

PR #2489, opened 2025-06-03
Add support for canned query parameters with defaults and longer descriptions. (Fixes #1258)

### LINKED ISSUE #1258: Canned queries that have named parameters should do something sensible when called without parameters.
If I call a canned query that includes named parameters, without passing any parameters, datasette runs the query anyway, resulting in an HTTP status code 400, and a visible error in the browser, with only a link back to home.  This means that one of the default links on https://site/database/ will lead to a broken page with no apparent way out.

image

Is there any way to skip performing the query when parameters aren't supplied, but otherwise render the usual canned query page?  Alternatively, can I supply default values for my parameters, either when defining my canned queries or when linking to the canned query page from the default database template.

etjones:
----
📚 Documentation preview 📚: https://datasette--2489.org.readthedocs.build/en/2489/

### DIFF
--- a/datasette/templates/query.html
+++ b/datasette/templates/query.html
@@ -57,7 +57,19 @@
             >
         {% endif %}
     {% endif %}
-    {% if named_parameter_values %}
+    
+    {# Debug lines removed to avoid Jinja2 Undefined errors #}
+
+    {% if canned_query and metadata and metadata.get('params') %}
+        <h3>Query parameters</h3>
+        {% for param in metadata['params'] %}
+            <p>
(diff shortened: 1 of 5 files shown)


etjones:
This is in response to #1258.

Canned queries have had this format: 
``` yaml
databases:
  fixtures:
    queries:
      neighborhood_search:
        params:
        - text
```
That's still the default, but this PR enables queries to have default values and tooltip descriptions as well: 
``` yaml
params:
- name: text
  default: ""
  description: "Longer description of the kind of text you want" 
```

Issue #2429, opened 2024-09-17
Datasette server redirects //example.com/ to https://example.com


psd:
The datasette server command 302 redirects paths such as `//example.com/` to '//example.com' which is followed by curl and browsers to 'https://example.com'.

Note, the path needs to begin with double-slash and end in a slash.

There's a log of how to reproduce the issue locally in this [gist](https://gist.github.com/psd/29834efeeb3d276b62a71f210989f1a9).

I've been unable to reproduce this issue on production servers, so could be something to flag more strongly in documentation if this is expected behaviour in development.

jamesjefferies:
We've now seen this on a Datasette on Heroku instance - i.e. on a production server 🤔

Wondering whether it is this redirect in the 404 handler? When there is a trailing `/` https://github.com/simonw/datasette/blob/7a602140df3646820adc45963daf7fc5dcd2a009/datasette/app.py#L1804

simonw:
I published a security advisory for this, no that both patched versions are released:
- https://github.com/simonw/datasette/security/advisories/GHSA-w832-gg5g-x44m

jamesjefferies:
Great, thank you @simonw - it is much appreciated!

PR #2500, opened 2025-09-11
Fix for open redirect - identified in Issue 2429

jamesjefferies:
The 404 processing ends up redirecting a request with multiple path slashes to that site, i.e.

https://my-site//shedcode.co.uk will redirect to https://shedcode.co.uk

Identified in issue #2429 

This commit uses a regular expression and a substitution to remove the multiple leading slashes before redirecting.

----
📚 Documentation preview 📚: https://datasette--2500.org.readthedocs.build/en/2500/


simonw:
Ouch! Thanks for the report. I'll look at this today.

jamesjefferies:
Have you had chance to have a look yet @simonw ?

simonw:
Thanks for this. I'm going to merge as-is, then move the fix to the `asgi_send_redirect()` function to reduce the change of the same bug affecting things from other places.

Issue #2528, opened 2025-10-24
Remove the old permission_allowed system and Permission class entirely


simonw:
It's time to ditch the old permissions system entirely and move everything to `datasette.allowed()` and friends.

This is why we fixed a bunch of plugins pinning alpha 1.0a19:
- #2516

simonw:
Told Claude:

> `Read https://github.com/simonw/datasette/issues/2528 and then prepare a detailed markdown report on what needs to be  done, put that in plan.md`

-----

# Plan: Remove Old Permission System (Issue #2528)

## Overview

Remove the legacy `permission_allowed` system and `Permission` class entirely from Datasette, consolidating all permission handling around the newer `datasette.allowed()` API and related functions.

**Related Issues:** #2516
**Breaking Change:** Yes
**Milestone:** Alpha with new permissions

## Current State Analysis

### Old System Components to Remove

1. **`datasette.permission_allowed()` method**
   - Old async method that returns boolean for permission checks
   - Used throughout codebase and by plugins
   - Need to identify all call sites

2. **`Permission` class**
   - Legacy dataclass/class representing a permission
   - Located in `datasette/permissions.py` or similar
   - May have methods and properties that need migration

3. **Related hooks/APIs**
   - `permission_allowed` plugin hook (if exists)
   - Any middleware or decorators using old system
   - Template helpers/functions

### New System (Replacement)

1. **`datasette.allowed()` method**
   - Returns Resource objects or list of allowed resources
   - Works with Action objects
   - Supports bulk checking via `allowed_resources()`

2. **`datasette.allowed_resources()` method**
   - Bulk permission checking
   - Returns list of Resource objects
   - Already implemented and working

3. **`Action` dataclass**
   - Defines actions with resource classes
   - Supports `also_requires` for dependencies
   - Located in `datasette/permissions.py`

4. **`Resource` class hierarchy**
   - `DatabaseResource`, `TableResource`, etc.
   - Provides typed resources with parent/child
   - Used throughout new system

## Migration Tasks

### Phase 1: Inventory and Analysis

- [ ] **Find all `permission_allowed()` call sites**
  - Search codebase for `permission_allowed(`
  - Search for `await.*permission_allowed`
  - Document each usage with context

- [ ] **Find all `Permission` class usages**
  - Search for `Permission(` instantiation
  - Search for `isinstance(.*Permission)`
  - Search for imports of Permission class

- [ ] **Identify affected plugins**
  - Check which core plugins use old system
[...]

PR #2535, opened 2025-10-25
Clean up obsolete permission code in favor of new actions and .allowed() method

simonw:
Refs:
- #2528

----
📚 Documentation preview 📚: https://datasette--2535.org.readthedocs.build/en/2535/

### DIFF
--- a/tests/test_permission_endpoints.py
+++ b/tests/test_permission_endpoints.py
@@ -166,9 +166,40 @@ async def test_allowed_json_pagination():
 
 
 @pytest.mark.asyncio
-async def test_allowed_json_total_count(ds_with_permissions):
+async def test_allowed_json_total_count(tmp_path_factory):
     """Test that /-/allowed.json returns correct total count."""
-    response = await ds_with_permissions.client.get("/-/allowed.json?action=view-table")
+    from datasette.database import Database
+
+    # Use temporary file databases to avoid leakage from other tests
+    tmp_dir = tmp_path_factory.mktemp("test_allowed_json_total_count")
+
+    ds = Datasette()
+    await ds.invoke_startup()
+
+    # Create test databases with tables
+    analytics_db = ds.add_database(
+        Database(ds, path=str(tmp_dir / "analytics.db")), name="analytics"
+    )
+    await analytics_db.execute_write(
+        "CREATE TABLE IF NOT EXISTS users (id INTEGER PRIMARY KEY, name TEXT, email TEXT)"
+    )
+    await analytics_db.execute_write(
+        "CREATE TABLE IF NOT EXISTS events (id INTEGER PRIMARY KEY, event_type TEXT, user_id INTEGER)"
+    )
+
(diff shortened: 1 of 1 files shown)


datasette/templates/debug_check.html
```diff
@@ -110,7 +110,7 @@ <h1>Permission Check</h1>
         <label for="action">Action (permission name):</label>
         <select id="action" name="action" required>
             <option value="">Select an action...</option>
-            {% for permission_name in sorted_permissions %}
+            {% for permission_name in sorted_actions %}
```
simonw: Let's do `action_name` here and in `debug_rules.html` too.

datasette/views/special.py
```diff
@@ -140,42 +144,43 @@ async def get(self, request):
...
+        # Convert to Resource object
+        if resource_1 and resource_2:
+            resource_obj = TableResource(database=resource_1, table=resource_2)
+            resource_for_response = (resource_1, resource_2)
+        elif resource_1:
+            resource_obj = DatabaseResource(database=resource_1)
+            resource_for_response = resource_1
+        else:
+            resource_obj = InstanceResource()
+            resource_for_response = None
```
simonw: This looks wrong. We can tell which class we should be using by looking up the action by name in `ds.actions[action_name]` and then accessing its `resource_class` property.

datasette/views/special.py
```diff
@@ -667,16 +680,12 @@ async def shared(self, request):
...
-                for key, value in self.ds.permissions.items()
-                if value.takes_database
+                key for key, value in self.ds.actions.items() if value.takes_database
             ],
             "resource_permissions": [
-                key
-                for key, value in self.ds.permissions.items()
-                if value.takes_resource
+                key for key, value in self.ds.actions.items() if value.takes_resource
             ],
```
simonw: Call these `"child_actions"`.

datasette/app.py
```diff
@@ -1450,20 +1228,88 @@ async def allowed(
...
+            else:
+                old_style_resource = None
+
+            # If restrictions don't allow this action, deny it
+            if not restrictions_allow_action(
+                self, actor["_r"], action, old_style_resource
+            ):
+                result = False
+
+        # Log the permission check for debugging
```
simonw: Instead of storing a `"resource"` key, lets store a `"parent"` and `"child"` key, both of which are string or None. In fact lets start using a dataclass for the things that go in `self._permission_checks` just so we get types in there.

tests/test_special.py
```diff
@@ -0,0 +1,196 @@
+"""
```
simonw: Rename this file from `test_special.py` to `test_search_tables.py`

simonw:
Told Claude to address my comments on this PR: https://gistpreview.github.io/?f0fd75de00a4599167e3003676bc6b06

simonw:
Now that I fixed all of the xfail tests in:
- #2538 

I think this is safe to land!

Issue #2560, opened 2025-10-31
Add pagination to internal datasette.allowed_resources() method


simonw:
> The lack of pagination here is worrying. Also it's weird that this one has a include_is_private when there is a whole other mechanism to get the reasons

_Originally posted by @simonw in https://github.com/simonw/datasette/pull/2551#discussion_r2464430452_

simonw:
This method here: https://github.com/simonw/datasette/blob/e4be95b16c99dd8d49fc6d89a684764bf731f1d9/datasette/app.py#L1143-L1199

It currently returns `list["Resource"]` - but the whole point of the new SQL resources system is to efficiently support potentially thousands of resources. Having a method that returns everything undermines that goal.

simonw:
Claude Code transcript: https://gistpreview.github.io/?6ea0291dbbf5e0456a09c276ae8f6b36

PR #2562, opened 2025-10-31
Add keyset pagination to allowed_resources()

simonw:
This replaces the unbounded list return with PaginatedResources, which supports efficient keyset pagination for handling thousands of resources.

Refs:
- #2560

Changes:
- allowed_resources() now returns PaginatedResources instead of list
- Added limit (1-1000, default 100) and next (keyset token) parameters
- Added include_reasons parameter (replaces allowed_resources_with_reasons)
- Removed allowed_resources_with_reasons() method entirely
- PaginatedResources.all() async generator for automatic pagination
- Uses tilde-encoding for tokens (matching table pagination)
- Updated all callers to use .resources accessor
- Updated documentation with new API and examples

The PaginatedResources object has:
- resources: List of Resource objects for current page
- next: Token for next page (None if no more results)
- all(): Async generator that yields all resources across pages

Example usage:
    page = await ds.allowed_resources("view-table", actor, limit=100)
    for table in page.resources:
        print(table.child)

    # Iterate all pages automatically
    async for table in page.all():
        print(table.child)

----
📚 Documentation preview 📚: https://datasette--2562.org.readthedocs.build/en/2562/

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -458,9 +458,9 @@ Example usage:
     for table in page.resources:
         print(f"{table.child}: {table.reasons}")
 
-The ``page.all()`` async generator automatically handles pagination under the hood, fetching additional pages as needed and yielding all resources one at a time.
+The ``page.all()`` async generator automatically handles pagination, fetching additional pages and yielding all resources one at a time.
 
-This method uses :ref:`datasette_allowed_resources_sql` under the hood and is an efficient way to list the databases, tables or queries visible to a user.
+This method uses :ref:`datasette_allowed_resources_sql` under the hood and is an efficient way to list the databases, tables or other resources that an actor can access for a specific action.
 
 .. _datasette_allowed_resources_sql:
 


simonw:
Tested this manually too:
```bash
uv run python -m asyncio
```
Then:
```python
ds = Datasette(["manytables.db"])
await ds.invoke_startup()
await ds.client.get("/-/versions")

page = await ds.allowed_resources("view-table")
for table in page.resources:
    print(table.child)

# Iterate all pages automatically
async for table in page.all():
    print(table.child)
```

Here's the `manytables.db` file I am using for tests: https://static.simonwillison.net/static/2025/manytables.db

datasette/views/special.py
```diff
@@ -268,19 +268,38 @@ async def _allowed_payload(self, request, has_debug_permission):
...
-                    parent=parent_filter,
-                )
+            allowed_rows = []
+            result = await self.ds.allowed_resources(
+                action=action,
+                actor=actor,
+                parent=parent_filter,
+                include_reasons=has_debug_permission,
+            )
+            async for resource in result.all():
```
simonw: Need to review this.

simonw:
I'll address the expensive uses of `.all()` in later issues.

Issue #2572, opened 2025-11-02
Fix problems with actor restrictions


simonw:
I'm not yet fully comfortable with how actor restrictions work - the bit where the code adds a bunch of extra deny rules looks complex and potentially error-prone to me.

simonw:
On digging in yeah, there's a bug in there: https://gistpreview.github.io/?7d170f364d7d24839b69c006e7e0ba39

Relevant extract from that report:

> **Setup:**
> - Config grants access to `database1` at the database level: `databases.database1.allow: {id: "user"}`
> - Actor has table-level restrictions allowing only `table2` and `table3`:
>   ```python
>   {"id": "user", "_r": {"r": {"database1": {"table2": ["vt"], "table3": ["vt"]}}}}
>   ```
> 
> **Expected Behavior:**
> - ✅ `table2` and `table3` should be accessible (in restriction allowlist)
> - ❌ `table1` and `table4` should be DENIED (not in restriction allowlist)
> 
> **Actual Behavior:**
> - ✅ `table2` and `table3` are accessible
> - 🐛 `table1` and `table4` are ALSO accessible (BUG!)

Full Markdown report it wrote: https://gist.github.com/simonw/0aab659147e39bea96f8d0db096bba5d

simonw:
I think there may be a better solution. Right now we have allow/deny rules and a rule that says more specific (e.g. child level)  beats less specific (parent or root level). But for the actor restrictions stuff that may not be enough - what if we introduce a 
fourth column that's a precedence column? Actor restriction rules can then take precedence over configuration rules.

This would be a  BIG change, changing the existing SQL contract such that PermissionSQL returns an additional column

Could it work though? And could it be done in a way that is easier for end developers to understand?

simonw:
Claude got excited about it, glazed me and said it was a "brilliant idea". https://gistpreview.github.io/?8325eac15114c7e5f3c8f2eb66de86e8

Interestingly it proposed NOT using an extra column, and instead doing this:

```python
PRECEDENCE_RESTRICTIONS = 0  # Actor restrictions (highest priority)
PRECEDENCE_NORMAL = 1        # Config & plugins
PRECEDENCE_DEFAULTS = 2      # Default allow rules (lowest priority)

@dataclass
class PermissionSQL:
    sql: str  # Still returns 4 columns: parent, child, allow, reason
    params: dict[str, Any] | None = None
    source: str | None = None
    precedence: int = 1  # NEW: Default to normal priority

[...]

PR #2573, opened 2025-11-03
New restrictions SQL mechanism

simonw:
Refs:
- #2572

----
📚 Documentation preview 📚: https://datasette--2573.org.readthedocs.build/en/2573/


simonw:
Coverage here didn't look good enough so I added more tests.

Issue #2579, opened 2025-11-05
Release 0.65.2


simonw:
Mainly this is to ship the open redirect fix from:
- #2429

But I also want to get things working with Python 3.14.

simonw:
I decided not to officially drop support for Python 3.9 because this has a security fix in and I want users stuck on that Python version to still be able to upgrade Datasette 0.x.

I tried and failed to get the CI tests passing on 3.9 so I ditched those, but my own manual testing on 3.9 locally appeared to work OK.

(Note that Python 3.9 is officially EOL as-of 31st October 2025: https://devguide.python.org/versions/)

simonw:
I'm going to fix the Cloud Run integration as part of this work too. If people have deployed Datasette to Cloud Run in the past they need to be able to update those deployments.

- #2511

simonw:
I'm taking the opportunity to clean up a few warnings, then I'm going to ship this.

simonw:
Claude Code session that fixed those warnings: https://gistpreview.github.io/?a75bd84e7864e178cfcfef8466c4c394

OK, release notes next: https://github.com/simonw/datasette/compare/0.65.1..0.65.x

simonw:
- Fixes an **open redirect** security issue: Datasette instances would redirect to `example.com/foo/bar` if you accessed the path `//example.com/foo/bar`. Thanks to [James Jefferies](https://github.com/jamesjefferies) for the fix. [#2429](https://github.com/simonw/datasette/issues/2429)
- Upgraded for compatibility with Python 3.14.
- Fixed `datasette publish cloudrun` to work with changes to the underlying Cloud Run architecture. [#2511](https://github.com/simonw/datasette/issues/2511)
- Minor upgrades to fix warnings, including `pkg_resources` deprecation.

Issue #2511, opened 2025-10-08
Error deploying to Cloud Run


simonw:
https://github.com/simonw/datasette/actions/runs/18358554568/job/52296691802

```
Error: google-github-actions/setup-gcloud failed with: failed to execute command `gcloud --quiet auth activate-service-account *** --key-file -`: ERROR: gcloud failed to load: module 'collections' has no attribute 'MutableMapping'
    gcloud_main = _import_gcloud_main()
    import googlecloudsdk.gcloud_main
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
    from googlecloudsdk.api_lib.util import exceptions
    from googlecloudsdk.core.resource import resource_printer
    from googlecloudsdk.core.resource import config_printer
    from googlecloudsdk.core.resource import resource_printer_base
    from googlecloudsdk.core.resource import resource_projector
    from google.protobuf import json_format as protobuf_encoding
    __import__('pkg_resources').declare_namespace(__name__)
    import packaging.requirements
    from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
    collections.MutableMapping.register(ParseResults)

This usually indicates corruption in your gcloud installation or problems with your Python interpreter.

Please verify that the following is the path to a working Python 2.7 or 3.5+ executable:
    /opt/hostedtoolcache/Python/3.10.18/x64/bin/python3

If it is not, please set the CLOUDSDK_PYTHON environment variable to point to a working Python 2.7 or 3.5+ executable.

If you are still experiencing problems, please reinstall the Cloud SDK using the instructions here:
    https://cloud.google.com/sdk/
```

simonw:
New error: https://github.com/simonw/datasette/actions/runs/18359225947/job/52298936998

> `error parsing HTTP 412 response body: invalid character 'C' looking for beginning of value: "Container Registry is deprecated and shutting down, please use the auto migration tool to migrate to Artifact Registry (gcloud artifacts docker upgrade migrate --projects='datasette-222320'). For more details see: https://cloud.google.com/artifact-registry/docs/transition/auto-migrate-gcr-ar"`

https://cloud.google.com/artifact-registry/docs/transition/auto-migrate-gcr-ar

simonw:
Working on this in a PR now:
- https://github.com/simonw/datasette/pull/2581

simonw:
Landed that on `main`, next step is backport it to 0.65.x.

simonw:
[...]

PR #2581, opened 2025-11-05
Get publish cloudrun working with latest Cloud Run

simonw:
Refs:
- #2511 
- #2579 

----
📚 Documentation preview 📚: https://datasette--2581.org.readthedocs.build/en/2581/

### DIFF
--- a/datasette/publish/cloudrun.py
+++ b/datasette/publish/cloudrun.py
@@ -23,7 +23,9 @@ def publish_subcommand(publish):
         help="Application name to use when building",
     )
     @click.option(
-        "--service", default="", help="Cloud Run service to deploy (or over-write)",
+        "--service",
+        default="",
+        help="Cloud Run service to deploy (or over-write)",
     )
     @click.option("--spatialite", is_flag=True, help="Enable SpatialLite extension")
     @click.option(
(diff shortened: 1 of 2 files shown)


simonw:
I'll get this working against `main` and then backport it to the `0.65.x` branch.

simonw:
I hit an error in this bit:
https://github.com/simonw/datasette/blob/fcae0aae6ea2ed3bc34ca62150147738bcacb279/datasette/publish/cloudrun.py#L289-L304

Some of my services have `"status"` but do not have `"url"` within that block.

It looks like those are old services that failed at some point. I'm going to filter them out.

simonw:
I ran a manual test of this with no `--service` option and it correctly showed me a list and let me set a name.

simonw:
This failed:
```bash
datasette publish cloudrun fixtures.db --service test-nov-5-2025
```
Because I upgraded the Docker image to use Python 3.14 but it defaults to attempting to the current live 0.65.1 version which is not compatible with Python 3.14.

simonw:
Running it like this DID work:
```bash
uv run datasette publish cloudrun fixtures.db --service test-nov-5-2025 --branch 0.65.x
```
I'm going to NOT ship the upgraded Python base image as part of the 0.65.x release. I'll bump that in `main` later.

Issue #2592, opened 2025-11-12
datasette --default-deny option


simonw:
I'm going to try adding this feature to see how it feels:
```bash
datasette --default-deny data.db --root
```
The `--default-deny` option would be passed as a keyword argument to the `Datasette(..., default_deny=True)` constructor, and set as `datasette.default_deny=True` on the instance.

It would then affect this code here:

https://github.com/simonw/datasette/blob/354d7a28732b701d5ebee334fc32a6e6e74ce0b2/datasette/default_permissions.py#L348-L367

If `default_deny` is set then those default rules are NOT added by that plugin hook implementation.

The goal of this feature is to make it as easy as possible for people to run a locked-down Datasette instance where only users who are authenticated using whatever extra auth mechanisms they configure can view anything.

I need this for Datasette Cloud.

simonw:
Initial prototype implementation by Claude Code: http://gistpreview.github.io/?b24a4ed0f3953708fbb2aa22d09dbb0f

Summary:

> Implementation Summary
> 
> 1\. CLI Option (datasette/cli.py)
> 
> - Added --default-deny flag to the serve command
> - The flag is passed through to the Datasette constructor
> 
> 2\. Core Implementation (datasette/app.py)
> 
> - Added default_deny parameter to Datasette.__init__() with default value False
> - Set self.default_deny instance attribute to track the setting
> 
> 3\. Permission Logic (datasette/default_permissions.py)
> 
> - Modified default_action_permissions_sql() function to check datasette.default_deny
> - When enabled, it skips returning default allow rules for:
>     - view-instance
>     - view-database
>     - view-database-download
>     - view-table
>     - view-query
>     - execute-sql
> 
> 4\. Tests (tests/test_default_deny.py)
> 
> Created comprehensive tests covering:
> - Default behavior: anonymous users can access content normally
> - With --default-deny: anonymous users are denied access to all resources
> - Root user retains full access even with --default-deny enabled
> - Config allow blocks still work to grant specific permissions
> - Basic permission checks for all default actions
> 
> 5\. Documentation (docs/authentication.rst)
> 
> Added a new section "Denying all permissions by default" that explains:
> - The purpose and use cases for --default-deny
> - How to use it with examples
> - Interaction with root user and configuration rules
> - Example showing how to grant access to specific authenticated users
> 
> Usage Example
> ```
> # Run Datasette with all access denied by default
> datasette --default-deny data.db --root
> 
> # With configuration to allow specific users
[...]

PR #2593, opened 2025-11-12
--default-deny option

simonw:
Refs:
- #2592

I'm not going to merge this until I've deployed it somewhere and exercised it and made sure it feels right.

----
📚 Documentation preview 📚: https://datasette--2593.org.readthedocs.build/en/2593/

### DIFF
--- a/datasette/cli.py
+++ b/datasette/cli.py
@@ -440,7 +440,7 @@ def uninstall(packages, yes):
 )
 @click.option(
     "--default-deny",
-    help="Deny all permissions by default, requiring explicit grants",
+    help="Deny all permissions by default",
     is_flag=True,
 )
 @click.option(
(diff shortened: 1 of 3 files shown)


simonw:
I manually tested this like so:
```bash
uv run --with-editable . \
  --with datasette-unsafe-actor-debug datasette \
  --default-deny \
  data.db \
  --config datasette.yaml \
  -s plugins.datasette-unsafe-actor-debug.enabled 1
```
Where `datasette.yaml` had this:

```yaml
allow:
  id: alice
```
Then I visited http://127.0.0.1:8001/-/unsafe-actor and confirmed I could sign in as `{"id": "alice"}` and see stuff but any other user or anonymous did not work.

simonw:
Deployed and tested this on Datasette Cloud, it's good!

Issue #2602, opened 2025-11-27
Refactor default_permissions.py into a package


simonw:
The `default_permissions.py` file is a bit large and hard to follow:

https://github.com/simonw/datasette/blob/68f1179bac991b5e37b99a5482c40134f317c04f/datasette/default_permissions.py#L1-L494

 494 lines.

simonw:
Suggested by Claude Opus 4.5 as part of this ad-hoc code review: https://github.com/simonw/research/tree/main/datasette-sql-permissions-review#611-split-default_permissionspy

PR #2603, opened 2025-11-27
Split default_permissions.py into a package

simonw:
Refs:
- #2602

Claude Code: https://gistpreview.github.io/?60291b48ed4862d5edbd046fefc5d238

----
📚 Documentation preview 📚: https://datasette--2603.org.readthedocs.build/en/2603/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,19 +27,6 @@ jobs:
       run: |
         pip install -e '.[test]'
         pip freeze
-    - name: Debug hooks
-      run: |
-        python -c "
-from datasette import plugins
-import datasette.default_permissions as dp
-print('Module file:', dp.__file__)
-print('Hooks registered:')
-for c in plugins.pm.get_hookcallers(dp):
-    print(f'  {c.name}')
-print('permission_resources_sql implementations:')
-for impl in plugins.pm.hook.permission_resources_sql.get_hookimpls():
-    print(f'  {impl.function.__name__} from {impl.plugin_name}')
-"
     - name: Run tests
       run: |
         pytest -n auto -m "not serial"


datasette/default_permissions/tokens.py
```diff
@@ -0,0 +1,114 @@
...
+if TYPE_CHECKING:
+    from datasette.app import Datasette
+
+import itsdangerous
+
+from datasette import hookimpl
+
+
+@hookimpl
+def actor_from_request(datasette: "Datasette", request) -> Optional[dict]:
```
simonw: Rename this to a more clear function name.

datasette/default_permissions/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    "default_allow_sql_check",
+    "default_action_permissions_sql",
+    "actor_from_request",
+    "skip_csrf",
+    "canned_queries",
+    # Utility functions and classes
+    "restrictions_allow_action",
+    "ActorRestrictions",
+    "DEFAULT_ALLOW_ACTIONS",
+]
```
simonw: don't need this

simonw:
The flaky test failures are really baffling, and I cannot replicate them locally.

simonw:
OK, it took a lot of poking around but it looks like the tests fail on `pull_request` - which tests against a merge commit to `main` - but pass on `test`.

So it's the merge from main that introduces the bug.

Claude Code says:

> ⏺ Your branch has the DELETE FROM catalog_databases line removed, but the merge commit has it present (from main).
> 
>   This means your branch diverged from main after this line was removed, but main still has it. The merge is keeping the main version.
> 
>   This could be the cause! If catalog_databases is being cleared improperly, it could affect what databases are visible.
> 
>   Let me check if this is actually the issue by testing:
> 
> ⏺ Bash(git checkout test-merge-commit && git show sql-permissions-refactor:datasette/utils/internal_db.py > datasette/utils/internal_db.py && source                     timeout: 1m 
>       /tmp/datasette-tes…)                                                                                                                                               0s
>   ⎿  collecting ... collected 1 item                                                
> 
>      tests/test_api.py::test_homepage PASSED                                  [100%]
>      … +3 lines (ctrl+o to expand)
> 
> ⏺ FOUND IT! The issue is the DELETE FROM catalog_databases line in internal_db.py. When that line is present (from main), it deletes database entries from the catalog, causing
>   databases to not be visible.
> 
>   Your branch has that line removed, but when GitHub merges for the PR, it picks up main's version which still has the line.
> 
>   You need to ensure your branch has internal_db.py rebased onto main, or the change to remove that line needs to be on main first.
>

simonw:
... or maybe not? Merging from main doesn't seem to fix the bug. Still digging.

simonw:
... after all of this it turns out there's a bug on `main` right now! I'll fix that and come back to this:
- https://github.com/simonw/datasette/issues/2606

Issue #2473, opened 2025-03-19
Error 500 no such column: "geometry_columns" - should this be a string literal in single-quotes?


jpmorr:
I'm trying to test datasette on one of my own databases to do some exploration but the interface doesn't load and just displays the Error 500 message given above. 

The traceback in the terminal is as follows:
```python
OperationalError: no such column: "geometry_columns" - should this be a string literal in single-quotes?
Traceback (most recent call last):
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\site-packages\datasette\app.py", line 1354, in route_path
    response = await view(request, send)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\site-packages\datasette\views\base.py", line 134, in view
    return await self.dispatch_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\site-packages\datasette\views\base.py", line 91, in dispatch_request
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\site-packages\datasette\views\index.py", line 34, in get
    hidden_table_names = set(await db.hidden_table_names())
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\site-packages\datasette\database.py", line 401, in hidden_table_names
    has_spatialite = await self.execute_fn(detect_spatialite)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\site-packages\datasette\database.py", line 213, in execute_fn
    return await asyncio.get_event_loop().run_in_executor(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\xxx\anaconda3\envs\new_base\Lib\concurrent\futures\thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #2001, opened 2023-01-23
Datasette is not compatible with SQLite's strict quoting compilation option


gwk:
I have linked Python3.11 on macOS against recent SQLite that was compiled using `-DSQLITE_DQS=0`. This option disables interpretation of double-quoted identifiers as string literals, described in the SQLite docs as a "MySQL 3.x misfeature". See https://www.sqlite.org/quirks.html#dblquote for background.

Datasette uses the double-quote syntax in a number of key places, and is thus completely broken in this environment.

My experience was to `pip install datasette`, then run `datasette serve -I my-data.db`. When I visit `http://127.0.0.1:8001` I get a 500 response.

The error: `sqlite3.OperationalError: no such column: geometry_columns`

The responsible SQL: `'select 1 from sqlite_master where tbl_name = "geometry_columns"'`

I then installed datasette from GitHub master in development mode and changed the offending SQL to use correct quotes: `"select 1 from sqlite_master where tbl_name = 'geometry_columns'"`.

With this change, I get a little further, but have the same problem with the first table name in my database (in my case, "Meta"):
```
OperationalError: no such column: Meta
Traceback (most recent call last):
  File "/Users/gwk/external/datasette/datasette/app.py", line 1522, in route_path
    response = await view(request, send)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gwk/external/datasette/datasette/views/base.py", line 151, in view
    return await self.dispatch_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gwk/external/datasette/datasette/views/base.py", line 105, in dispatch_request
    response = await handler(request)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gwk/external/datasette/datasette/views/index.py", line 70, in get
    "fts_table": await db.fts_table(table),
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gwk/external/datasette/datasette/database.py", line 363, in fts_table
    return await self.execute_fn(lambda conn: detect_fts(conn, table))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gwk/external/datasette/datasette/database.py", line 213, in execute_fn
    return await asyncio.get_event_loop().run_in_executor(
[...]

PR #2616, opened 2025-12-16
Fix compatibility with SQLite built with the recommended -DSQLITE_DQS=0 option

tuffnatty:
This PR addresses the problem in #2001 and #2473, which is also touched by #2004 and #2474.

The failing tests are failing in the current main as well.

----
📚 Documentation preview 📚: https://datasette--2616.org.readthedocs.build/en/2616/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1041,12 +1041,12 @@ class Datasette:
             for db_name, db in self.databases.items():
                 if count >= SQLITE_LIMIT_ATTACHED or db.is_memory:
                     continue
-                sql = 'ATTACH DATABASE "file:{path}?{qs}" AS [{name}];'.format(
+                sql = "ATTACH DATABASE ? AS {};".format(escape_sqlite(db_name))
+                location = "file:{path}?{qs}".format(
                     path=db.path,
                     qs="mode=ro" if db.is_mutable else "immutable=1",
-                    name=db_name,
                 )
-                conn.execute(sql)
+                conn.execute(sql, [location])
                 count += 1
 
     def add_message(self, request, message, type=INFO):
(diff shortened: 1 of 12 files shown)


tuffnatty:
One of these changes probably needs some discussion. It's https://github.com/simonw/datasette/pull/2616/commits/c6a47e825a569e7605f1346a3d8fb8b098b0f4e7.

Before disabling double-quoted string literals, passing extra query string parameters not matching any known column name would add always-false WHERE clauses to SQL, e.g.: `&z=baz` would generate `WHERE "z" = 'baz'`, which, when double-quoted string literals are enabled, is an always-false string literal comparison, and the view returns an empty result set. But when double-quoted strings are disabled, this WHERE clause generates an error.

Here I have chosen to ignore query string parameters not matching any known column. It changes the existing behaviour. Another option is to return an error. I am not sure returning an empty result set is a behaviour worth keeping.

tuffnatty:
I have switched from compiling SQLite3 with `-DSQLITE_DQS=0` to disabling double-quoted strings programmatically with `Connection.setconfig()` in tests on supported platforms (SQLite 3.29.0+, Python 3.12+). This is enough for testing, allows offline testing without rebuilding SQLite, and can also be easily backported to 0.65.x.

tuffnatty:
Hi @simonw ,

should I rebase this PR on the master branch fixing the new conflicts and keep supporting it, or should I bury hopes that one day datasette will switch to standard string quoting? More and more distributions switch to compiling SQLite with `-DSQLITE_DQS=0` (which is sane, standard-compliant and recommended by the upstream). My personal point of concern is FreeBSD binary packages. But if you have some reasons to keep using double quotes for string quoting throughout the code, I'll just keep using my fork and forget about this PR and #2617.

Looking forward to hearing from you.

tuffnatty:
Rebased against main. Fixed some new AI-generated code. Fixed SQL fragments in JSON API documentation.

Issue #2632, opened 2026-01-29
Flaky test on SQLite 3.25.3


simonw:
This flaky test only affects SQLite 3.25.3 and only for Python 3.10 and 3.12:

```
_______________ test_upsert[True-initial0-input0-expected_rows0] _______________
[gw1] linux -- Python 3.10.19 /opt/hostedtoolcache/Python/3.10.19/x64/bin/python

ds_write = <datasette.app.Datasette object at 0x7ff30b42a5c0>
initial = {'pk': 'id', 'rows': [{'id': 1, 'title': 'One'}], 'table': 'upsert_test'}
input = {'return': True, 'rows': [{'id': 1, 'title': 'Two'}]}
expected_rows = [{'id': 1, 'title': 'Two'}], should_return = True

    @pytest.mark.asyncio
    @pytest.mark.parametrize(
        "initial,input,expected_rows",
        (
            (
                # Simple primary key update
                {"rows": [{"id": 1, "title": "One"}], "pk": "id"},
                {"rows": [{"id": 1, "title": "Two"}]},
                [
                    {"id": 1, "title": "Two"},
                ],
            ),
            (
                # Multiple rows update one of them
                {
                    "rows": [{"id": 1, "title": "One"}, {"id": 2, "title": "Two"}],
                    "pk": "id",
                },
                {"rows": [{"id": 1, "title": "Three"}]},
                [
                    {"id": 1, "title": "Three"},
                    {"id": 2, "title": "Two"},
                ],
            ),
            (
                # rowid update
                {"rows": [{"title": "One"}]},
                {"rows": [{"rowid": 1, "title": "Two"}]},
                [
                    {"rowid": 1, "title": "Two"},
                ],
            ),
            (
                # Compound primary key update
                {"rows": [{"id": 1, "title": "One", "score": 1}], "pks": ["id", "score"]},
                {"rows": [{"id": 1, "title": "Two", "score": 1}]},
                [
                    {"id": 1, "title": "Two", "score": 1},
                ],
            ),
            (
                # Upsert with an alter
                {"rows": [{"id": 1, "title": "One"}], "pk": "id"},
                {"rows": [{"id": 1, "title": "Two", "extra": "extra"}], "alter": True},
                [{"id": 1, "title": "Two", "extra": "extra"}],
            ),
        ),
    )
    @pytest.mark.parametrize("should_return", (False, True))
    async def test_upsert(ds_write, initial, input, expected_rows, should_return):
[...]

PR #2633, opened 2026-01-29
Workaround for bug #2632

simonw:
Refs:

- #2632 

----
📚 Documentation preview 📚: https://datasette--2633.org.readthedocs.build/en/2633/


simonw:
Hard to say for sure if this fixes the intermittent bug because the bug is intermittent! I'm going to land it and hope it does the job.

PR #2636, opened 2026-02-09
write_wrapper plugin hook for intercepting write operations

### LINKED ISSUE #2637: write_wrapper plugin hook
This came up while working on `datasette-sqlite-history-json`:

- https://github.com/datasette/datasette-sqlite-history-json/pull/2

The desire was to run some custom SQL before and after any write operations, depending on the current request and actor.

Claude helpfully wrote a design document for this which made a solid case for a new plugin hook that wraps writes to the database: https://github.com/datasette/datasette-sqlite-history-json/blob/38360848cc0458ca339c77a92de98ef25f84ec3b/docs/design-change-groups.md

simonw:
> Fetch https://raw.githubusercontent.com/datasette/datasette-sqlite-history-json/38360848cc0458ca339c77a92de98ef25f84ec3b/docs/design-change-groups.md to /tmp using curl, then read it
> 
> Implement the proposed wrap_write() plugin hook, but with a difference. The proposal in that document has it returning a wrapper function. Instead, have it return a function that acts as a context manager - so the user provided function takes a connection and can then run some code, then yield to allow Datasette to execute its default write stuff, and then run more code afterwards.
> 
> Ask clarifying questions and propose your design before implementing anything

## Summary

This PR introduces a new `wrap_write` plugin hook that allows plugins to intercept and wrap database write operations. This enables use cases like transaction management, change tracking, audit logging, and other cross-cutting concerns around writes.

## Key Changes

- **New `wrap_write` hookspec** in `datasette/hookspecs.py`: Defines the plugin hook interface that accepts `datasette`, `database`, `request`, and `transaction` parameters and returns a generator-based context manager wrapper.

- **Request parameter propagation**: Added optional `request` parameter to `execute_write()`, `execute_write_script()`, `execute_write_many()`, and `execute_write_fn()` methods to pass HTTP request context to plugins.

- **Hook wrapping implementation** in `datasette/database.py`:
  - New `_wrap_fn_with_hooks()` method that collects and applies all registered `wrap_write` wrappers
  - New `_apply_wrap_write()` utility function that properly handles generator-based context managers, including exception propagation and result passing via `.send()`

- **View layer updates**: Modified write operations in `datasette/views/database.py`, `datasette/views/row.py`, and `datasette/views/table.py` to pass the `request` object to write methods.

- **Comprehensive test coverage** in new `tests/test_wrap_write.py` with 13 test cases covering:
  - Before/after hook execution
  - Result passing through yield
  - Exception handling in generators
  - Multiple plugin nesting
  - Request and transaction parameter passing
  - Integration with API write operations
  - Real-world use case example (change group pattern)

## Implementation Details

The `wrap_write` hook uses Python generators as context managers. Plugins return a generator function that:
- Runs setup code before the `yield`
- Yields to allow the write to execute
- Runs cleanup code after the `yield`
- Can capture the write result via `result = yield`
- Can handle exceptions thrown into the generator

Multiple wrappers are nested in reverse order (LIFO) so the first registered plugin is the outermost wrapper, maintaining intuitive execution order.

https://claude.ai/code/session_019hWsbmHqitdVNCNChN5aNJ

----
📚 Documentation preview 📚: https://datasette--2636.org.readthedocs.build/en/2636/

https://gisthost.github.io/?c4c12079434e69677e4aa8ac664b21b8/index.html

Refs:
- #2637

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -130,25 +130,27 @@ class Database:
         for connection in self._all_file_connections:
             connection.close()
 
-    async def execute_write(self, sql, params=None, block=True):
+    async def execute_write(self, sql, params=None, block=True, request=None):
         def _inner(conn):
             return conn.execute(sql, params or [])
 
         with trace("sql", database=self.name, sql=sql.strip(), params=params):
-            results = await self.execute_write_fn(_inner, block=block)
+            results = await self.execute_write_fn(
+                _inner, block=block, request=request
+            )
         return results
 
-    async def execute_write_script(self, sql, block=True):
+    async def execute_write_script(self, sql, block=True, request=None):
         def _inner(conn):
             return conn.executescript(sql)
 
         with trace("sql", database=self.name, sql=sql.strip(), executescript=True):
             results = await self.execute_write_fn(
-                _inner, block=block, transaction=False
+                _inner, block=block, transaction=False, request=request
             )
         return results
 
-    async def execute_write_many(self, sql, params_seq, block=True):
+    async def execute_write_many(self, sql, params_seq, block=True, request=None):
         def _inner(conn):
             count = 0
 
@@ -163,7 +165,9 @@ class Database:
         with trace(
             "sql", database=self.name, sql=sql.strip(), executemany=True
         ) as kwargs:
-            results, count = await self.execute_write_fn(_inner, block=block)
+            results, count = await self.execute_write_fn(
+                _inner, block=block, request=request
+            )
             kwargs["count"] = count
         return results
 
@@ -187,7 +191,8 @@ class Database:
             # Threaded mode - send to write thread
             return await self._send_to_write_thread(fn, isolated_connection=True)
 
-    async def execute_write_fn(self, fn, block=True, transaction=True):
+    async def execute_write_fn(self, fn, block=True, transaction=True, request=None):
+        fn = self._wrap_fn_with_hooks(fn, request, transaction)
         if self.ds.executor is None:
             # non-threaded mode
             if self._write_connection is None:
@@ -202,6 +207,25 @@ class Database:
             return await self._send_to_write_thread(
                 fn, block=block, transaction=transaction
             )
+
+    def _wrap_fn_with_hooks(self, fn, request, transaction):
+        from .plugins import pm
+
+        wrappers = pm.hook.wrap_write(
+            datasette=self.ds,
+            database=self.name,
+            request=request,
+            transaction=transaction,
+        )
+        wrappers = [w for w in wrappers if w is not None]
+        if not wrappers:
+            return fn
+        # Build the wrapped fn by nesting context manager generators.
+        # The first wrapper returned by pluggy is outermost.
+        original_fn = fn
+        for wrapper_factory in reversed(wrappers):
+            original_fn = _apply_wrap_write(original_fn, wrapper_factory)
+        return original_fn
 
     async def _send_to_write_thread(
         self, fn, block=True, isolated_connection=False, transaction=True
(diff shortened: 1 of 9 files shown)


simonw:
I pitched Claude the idea of having the plugin function able to `yield conn` with a completely new connection which would then be used by the rest of the stack, to allow for fancy ideas like having per-actor connections that persist across requests, or something like https://github.com/datasette/datasette-transactions

Quoting the response in full because it's worth recording somewhere:

> This is a fascinating idea. Let me think through it carefully.
> 
> * * * * *
> 
> `yield conn` --- Connection Substitution via `wrap_write`
> -------------------------------------------------------
> 
> The idea: if the generator yields a connection object instead of `None`, Datasette uses that yielded connection for the write instead of the default shared write connection. This would let a plugin:
> 
> 1.  **Hold open a dedicated connection per actor** --- reused across requests
> 2.  **Hold a transaction open across multiple HTTP requests** --- exactly what datasette-transactions does, but without needing to bypass `execute_write_fn` entirely
> 
> ### How datasette-transactions works today
> 
> The plugin completely sidesteps Datasette's write system. It:
> 
> -   Opens its own `sqlite3.connect(db.path, isolation_level=None, check_same_thread=False)` connections
> -   Manages them in a global `TransactionManager` singleton
> -   Registers its own routes (`/-/transactions/begin/...`, `/-/transactions/{uuid}`, etc.)
> -   Never touches `execute_write_fn` at all --- it runs SQL directly against its own connections
> 
> This means **writes done through datasette-transactions don't trigger any of the normal Datasette write hooks** --- no `wrap_write`, no `track_event`. They're invisible to the rest of the plugin ecosystem.
> 
> ### What `yield conn` could enable
> 
> If `wrap_write` allowed yielding a substitute connection:
> 
> ```
> @hookimpl
> def wrap_write(datasette, database, request):
>     # Look up whether this actor has an open transaction
>     tx_state = get_transaction_for_actor(request)
>     if tx_state is None:
>         return None  # Use default connection
> 
>     def wrapper(conn):
>         # Yield a different connection --- this one has an open transaction
>         yield tx_state.connection
> 
>     return wrapper
> 
> ```
> 
> Then writes going through the normal Datasette API (row insert/update/delete/drop, canned queries) would **automatically** be redirected to the actor's dedicated transactional connection. This would mean:
> 
> -   The existing Datasette table UI and JSON API would just work inside transactions
> -   Other `wrap_write` plugins (audit, change groups, authorizer) would still fire
> -   `track_event` would still fire
> -   No need for separate routes or a custom SQL execution endpoint
> 
> ### The tricky parts
> 
> **1\. The write thread model breaks down.** Today, `execute_write_fn` serializes everything through one connection on one dedicated thread. If a plugin yields a different connection, that connection might be owned by a different thread, or might need to live longer than a single `execute_write_fn` call. The yielded connection would need `check_same_thread=False` and its own locking.
> 
> **2\. Transaction boundary management.** The `with conn:` transaction wrapper in `_execute_writes` and the non-threaded path would need to be skipped or handled differently when a substitute connection is in use --- the plugin is managing the transaction itself. The `transaction` parameter being passed to the hook helps here: the plugin can see it and the hook could signal "I'm managing transactions myself."
> 
> **3\. Connection lifetime.** The substitute connection outlives a single write call. Who closes it? The plugin would need to manage lifecycle (begin, commit/rollback, close) through its own endpoints, just like datasette-transactions does today --- but the *execution* of writes flows through Datasette's normal path.
> 
> **4. `prepare_connection` hooks.** Datasette calls `_prepare_connection` on the default write connection. A substitute connection from a plugin wouldn't have had those hooks applied unless the plugin explicitly calls `datasette._prepare_connection(conn, database)` or Datasette does it automatically.
> 
> ### Implementation sketch
> 
> The change to `_apply_wrap_write` would be small. Currently:
> 
> ```
> result = fn(conn)  # always uses the original conn
> 
> ```
> 
> With `yield conn` support:
> 
> ```
> yielded = next(gen)
> # If generator yielded a connection, use that instead
> effective_conn = yielded if yielded is not None else conn
> result = fn(effective_conn)
> 
> ```
> 
> But the harder question is whether to skip the outer transaction wrapping (`with conn:`) when a substitute connection is used. That probably needs to be a separate flag or convention.
> 
> ### Is this worth building now?
> 
> I think the concept is powerful and the implementation change to `_apply_wrap_write` is minimal. But the real complexity is in the *interaction* with the write thread and transaction management --- specifically, when a plugin yields a substitute connection:
> 
> -   Should `execute_write_fn` still send the task to the write thread? (Probably yes, to maintain serialization)
> -   Should the `with conn:` transaction wrapper use the substitute connection? (Probably not --- the plugin owns the transaction)
> -   Should `_prepare_connection` be called on substitute connections? (Probably --- but only once, not per-write)

simonw:
Here's the plugin documentation Claude wrote: https://github.com/simonw/datasette/blob/ce1b4e509b39ae9b643640a1995b67e549cc13ea/docs/plugin_hooks.rst#write_wrapperdatasette-database-request-transaction

simonw (commented):
After several iterations this all looks great to me.

PR #2648, opened 2026-02-20
Respect metadata-defined facet ordering in sorted_facet_results

### LINKED ISSUE #2647: If facets are defined in metadata/settings use that defined order
Suggestion [from Discord](https://discord.com/channels/823971286308356157/823971286941302908/1474455540204834917):
```yaml
dbname:
  tables:
    table_name:
      facets:
         - region
         - provider
         - capability_1_score
         - capability_2_score
         - capability_3_score
```
Facets on the table page are currently ordered by the ones with the most results. We should change it so they use the defined `facets` order if one was defined, otherwise using the current behavior.

simonw:
> Suggestion [[from Discord](https://discord.com/channels/823971286308356157/823971286941302908/1474455540204834917)](https://discord.com/channels/823971286308356157/823971286941302908/1474455540204834917):
> ```yaml
> dbname:
>   tables:
>     table_name:
>       facets:
>       - region
>       - provider
>       - capability_1_score
>       - capability_2_score
> ```
> Facets on the table page are currently ordered by the ones with the most results. We should change it so they use the defined `facets` order if one was defined, otherwise using the current behavior.

(I pasted in the issue body text from #2647)

## Summary
This change ensures that facets defined in table metadata are returned in their configured order, rather than being sorted by result count and name. Request-based facets (added via `_facet` parameter) are still sorted by result count and appear after metadata-defined facets.

## Key Changes
- Modified `extra_sorted_facet_results()` in `datasette/views/table.py` to parse facet configurations from table metadata and maintain their defined order
- Metadata-defined facets are now sorted according to their order in the config, while request-based facets fall back to the original sorting behavior (by result count, then name)
- Added comprehensive test coverage in `tests/test_facets.py`:
  - Enhanced existing `test_other_types_of_facet_in_metadata()` to verify facets appear in metadata-defined order in HTML responses
  - Added new `test_metadata_facet_ordering()` to verify JSON response ordering for both metadata-only and mixed metadata/request-based facets

## Implementation Details
- The facet config parser handles both string format (`"state"`) and dict format (`{"date": "created"}`) for facet definitions
- Facets are split into two groups: those defined in metadata (sorted by config order) and those added via request parameters (sorted by result count)
- When no metadata facets are defined, the original sorting behavior is preserved for backward compatibility

https://claude.ai/code/session_01PbSHtjsUpNk3Fx7xjvVqDb

### DIFF
--- a/datasette/views/table.py
+++ b/datasette/views/table.py
@@ -1589,9 +1589,7 @@ async def table_view_data(
                     metadata_facet_names.append(fc)
                 elif isinstance(fc, dict):
                     metadata_facet_names.append(list(fc.values())[0])
-            metadata_order = {
-                name: i for i, name in enumerate(metadata_facet_names)
-            }
+            metadata_order = {name: i for i, name in enumerate(metadata_facet_names)}
             metadata_facets = []
             request_facets = []
             for f in extra_facet_results["results"].values():
--- a/datasette/views/table.py
+++ b/datasette/views/table.py
@@ -1580,11 +1580,37 @@ async def table_view_data(
         ]
 
     async def extra_sorted_facet_results(extra_facet_results):
-        return sorted(
-            extra_facet_results["results"].values(),
-            key=lambda f: (len(f["results"]), f["name"]),
-            reverse=True,
-        )
+        facet_configs = table_metadata.get("facets", [])
+        if facet_configs:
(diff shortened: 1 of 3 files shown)


simonw:
> Update the facets documentation to mention this

PR #2652, opened 2026-03-03
Apply `render_cell` plugin hook to primary key columns

westurner:
This update modifies the table view to execute the `render_cell` plugin hook against simple (single) primary key (or "Link") columns. Instead of always generating a default HTML anchor link pointing to the row view, Datasette now checks if any plugin returns a custom display value. If a custom rendering is provided, it is used; otherwise, it safely falls back to the default `<a href="...">` behavior.

Key Changes & Optimizations:
- Feature: Executes `render_cell` hooks for simple primary keys in `display_columns_and_rows`, safely resolving `pk_value` from the row.
- Performance: Pre-computes `has_single_pk`, `pk_index`, and `has_render_cell_hooks` outside of the O(N) row loop to eliminate redundant operations.
- Performance: Completely skips hook evaluation if no plugins are registered to listen to `render_cell`.
- Refactor: Streamlines row cell appending by using an inline `A if A is not None else B` fallback, which prevents duplicate dictionary initializations and safely handles intentional empty strings (`""`) returned by plugins.
- Compatibility: Passes `pks=pks_for_display` to the hook call to satisfy upstream API requirements.

Assisted By: Gemini 3.1 Pro

----
📚 Documentation preview 📚: https://datasette--2652.org.readthedocs.build/en/2652/


westurner:
From https://github.com/pulp for example:

> All contributors must indicate in the commit message of their contribution if they used AI to create them and the contributor is fully responsible for the content that they submit.

> This can be a label such as **Assisted By: <Tool>** or Generated by: <Tool> based on what was used.

So I added "Assisted By:". I would say this one was more of a pair coding with LLMs than a vibe coding. I ran the tests a number of times locally. I'm not sure why I have 15 local test failures before and after this PR. Does the Actions build run the tests with coverage?

datasette/views/table.py
```diff
@@ -195,22 +195,54 @@ async def display_columns_and_rows(
         for fk in await db.foreign_keys_for_table(table_name)
     }
 
+    has_single_pk = len(pks) == 1
+    has_render_cell_hooks = bool(pm.hook.render_cell.get_hookimpls())
```
westurner, line 689: It would be faster to check `has_render_cell_hooks = bool(pm.hook.render_cell.get_hookimpls())` outside of this function. Could there be a global bool for this? Is there a config reload?

datasette/views/table.py
```diff
@@ -195,22 +195,54 @@ async def display_columns_and_rows(
...
+                    value=pk_value,
+                    column=pks[0],
+                    table=table_name,
+                    pks=pks_for_display,
+                    database=database_name,
+                    datasette=datasette,
+                    request=request,
+                ):
+                    candidate = await await_me_maybe(candidate)
+                    if candidate is not None:
```
westurner, line 725: Should it be possible to set `plugin_display` to falsy values like `False` or `''` or `[]`?

```py
assert all(not bool(x) for x in [False, "", [], (), None])
```

westurner: should be checking for `is not None` instead of `not` now

datasette/views/table.py
westurner: So, this PR works but is slower than necessary

PR #2666, opened 2026-03-17
Fix startup hook to fire after metadata and schema tables are populated

### LINKED ISSUE #2665: Ensure metadata tables populated before plugin startup hook fires
Currently the `startup()` hook fires before the internal database tables have been populated from `metadata.yaml`.

It would be better if metadata populated first, because then plugins that have their own opinions about where metadata comes from could solve the challenge of merging or overriding default metadata themselves.

simonw:
> Currently the startup() hook fires before the u eternal database tables have been populated from metadata.yaml.
> 
> It would be better if metadata populated first, because then plugins that have their own opinions about where metadata comes from could solve the challenge of merging or overriding default metadata themselves.
>
> Run relevant tests with “uv run pytest -k” and then fix with ref/green TDD

> We should also ensure that the other internal tables from refresh schemas have been populated before that startup plugin hook fires

Previously, the startup() plugin hook fired before internal database
tables were populated from metadata.yaml and before catalog schema
tables were filled. This meant plugins couldn't read or modify metadata
during startup. Now invoke_startup() calls refresh_schemas() before
firing startup hooks, ensuring metadata and catalog tables are available.

https://claude.ai/code/session_01KL4t5FZYb32rZY7xaqrrZU

----
📚 Documentation preview 📚: https://datasette--2666.org.readthedocs.build/en/2666/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -697,7 +697,7 @@ class Datasette:
         ):
             await await_me_maybe(hook)
         # Ensure internal tables and metadata are populated before startup hooks
-        await self.refresh_schemas()
+        await self._refresh_schemas()
         for hook in pm.hook.startup(datasette=self):
             await await_me_maybe(hook)
         self._startup_invoked = True
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -696,6 +696,8 @@ class Datasette:
             env=self._jinja_env, datasette=self
         ):
             await await_me_maybe(hook)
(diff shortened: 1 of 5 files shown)


simonw:
Here's why this works:

https://github.com/simonw/datasette/blob/f0e543b8a94a90e4103773d1e95f5aa0edc7246f/datasette/app.py#L602-L607

PR #2667, opened 2026-03-17
Add column types system, refs #2664

### LINKED ISSUE #2664: Column types mechanism in Datasette core
For [datasette-files](https://github.com/simonw/datasette-files) I need a mechanism to specify which columns in a table reference files (text columns storing IDs like `df-12312412`), so that the UI that allows files to be attached to rows can kick in for that specific table in the correct place.

Other reasons column types could be useful:
- Specify a column as storing a geospatial point to support maps
- An `email` or `url` type which can help validate when data is edited
- BLOB columns that store a specific kind of data like JPEGs or PDFs
- Markdown columns which can then be rendered correctly when displayed

Introducing this into Datasette core makes sense as that way plugins can register additional column types and make use of the core concept in interesting ways.

simonw:
Implements the column types feature that lets Datasette and plugins annotate
columns with semantic types beyond SQLite storage types (e.g. markdown, email,
url, json, file, point). This enables type-appropriate rendering, validation,
form widgets, and API behavior.

- #2664

Key changes:
- New `column_types` internal DB table for storing assignments
- `ColumnType` dataclass in datasette/column_types.py with render_cell,
  validate, and transform_value methods
- `register_column_types` plugin hook for registering types
- Built-in url, email, and json column types
- Datasette API methods: get/set/remove_column_type(s),
  get_column_type_class
- Config loading from datasette.json `column_types` table config key
- `column_types` extra on the table JSON endpoint
- Column type info in display_columns extra
- Column type render_cell gets priority in rendering pipeline
- column_type/column_type_config args added to render_cell hookspec
- Write-path validation on insert and update

https://claude.ai/code/session_01SvPEPqHgURTWESRp28pTC3

----
📚 Documentation preview 📚: https://datasette--2667.org.readthedocs.build/en/2667/

### DIFF
--- a/tests/test_column_types.py
+++ b/tests/test_column_types.py
@@ -381,17 +381,6 @@ async def test_column_type_base_defaults():
     assert await ct.transform_value("val", None) == "val"
 
 
-@pytest.mark.asyncio
-async def test_column_type_with_config():
-    class TestType(ColumnType):
-        name = "test"
-        description = "Test type"
-
-    ct = TestType(config={"key": "value"})
-    assert ct.config == {"key": "value"}
-    assert ct.name == "test"
-
-
 # --- render_cell extra with column types ---
 
 
--- a/tests/test_column_types.py
+++ b/tests/test_column_types.py
@@ -1,7 +1,5 @@
 import logging
 
-import logging
-
 from datasette.app import Datasette
 from datasette.column_types import ColumnType
 from datasette.hookspecs import hookimpl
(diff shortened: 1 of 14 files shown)


tests/test_column_types.py
```diff
@@ -357,9 +375,21 @@ class TestType(ColumnType):
...
+
+@pytest.mark.asyncio
+async def test_column_type_with_config():
+    class TestType(ColumnType):
+        name = "test"
+        description = "Test type"
+
+    ct = TestType(config={"key": "value"})
+    assert ct.config == {"key": "value"}
+    assert ct.name == "test"
```
simonw: Pointless test.

PR #2678, opened 2026-03-24
Document call_with_supported_arguments as a supported public API

simonw:
Documenting these because I want to use the sync one in https://github.com/datasette/datasette-files

Mark both call_with_supported_arguments and async_call_with_supported_arguments
with the @documented decorator and add documentation to docs/internals.rst
so plugin authors can use these dependency injection utilities in their own code.

https://claude.ai/code/session_01DKogZpHwzCTrbeG4XjXmNc

----
📚 Documentation preview 📚: https://datasette--2678.org.readthedocs.build/en/2678/

### DIFF
--- a/datasette/utils/__init__.py
+++ b/datasette/utils/__init__.py
@@ -1086,12 +1086,35 @@ def _gather_arguments(fn, kwargs):
     return call_with
 
 
+@documented
 def call_with_supported_arguments(fn, **kwargs):
+    """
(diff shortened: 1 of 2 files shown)


simonw:
New docs start here: https://datasette--2678.org.readthedocs.build/en/2678/internals.html#call-with-supported-arguments-fn-kwargs

PR #2690, opened 2026-04-15
fix: harden sqlite identifier quoting

atharvasingh7007:
## Summary
- switch `escape_sqlite()` to standard SQLite double-quoted identifier escaping instead of bracket quoting
- replace the remaining raw `[{name}]` interpolations in Python query builders with `escape_sqlite()`
- add regressions for malicious table/database names in API, internal catalog, table counts, crossdb attach, and the helper PoC

## Root cause
`escape_sqlite()` wrapped identifiers in square brackets but did not safely handle `]` inside the identifier. SQLite does not provide a reliable bracket-escape form for embedded `]`, so a malicious table name could break out of that quoting.

This also left a few hand-written `[{name}]` call sites vulnerable even if the helper was corrected.

## Testing
- `python -m pytest tests/test_utils.py tests/test_api.py tests/test_internals_database.py tests/test_crossdb.py -k "escape_sqlite or closing_bracket or detect_fts_different_table_names"`
- `python -m ruff check datasette/app.py datasette/database.py datasette/facets.py datasette/filters.py datasette/utils/__init__.py datasette/utils/internal_db.py datasette/views/table.py tests/test_api.py tests/test_crossdb.py tests/test_internals_database.py tests/test_utils.py`

Closes #2677

----
📚 Documentation preview 📚: https://datasette--2690.org.readthedocs.build/en/2690/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -1225,10 +1225,10 @@ class Datasette:
             for db_name, db in self.databases.items():
                 if count >= SQLITE_LIMIT_ATTACHED or db.is_memory:
                     continue
-                sql = 'ATTACH DATABASE "file:{path}?{qs}" AS [{name}];'.format(
+                sql = 'ATTACH DATABASE "file:{path}?{qs}" AS {name};'.format(
                     path=db.path,
                     qs="mode=ro" if db.is_mutable else "immutable=1",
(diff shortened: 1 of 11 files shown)


simonw:
Handled in #2771

Issue #1752, opened 2022-05-28
Research if I can drop Janus


simonw:
> It seems to me Janus dependency is not necessary, `async with app.database_write_mutex(): out = await app.transaction(func)` may be enough.

Comment here: https://lobste.rs/s/fki4tj/architecture_notes_datasette#c_a2ihon

PR #2715, opened 2026-05-16
Replace Janus queue with asyncio.Future

simonw:
Closes:
- #1752

AI generated patch explanation: https://gisthost.github.io/?e2b8d9c7666e988b5c003ff5e5ef3098

----
📚 Documentation preview 📚: https://datasette--2715.org.readthedocs.build/en/2715/

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -4,7 +4,6 @@ from collections import namedtuple
 import inspect
 import os
 from pathlib import Path
-import janus
 import queue
 import sqlite_utils
 import sys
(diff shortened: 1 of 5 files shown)


datasette/database.py
```diff
@@ -390,18 +392,15 @@ async def _send_to_write_thread(
             )
             self._write_thread.start()
         task_id = uuid.uuid5(uuid.NAMESPACE_DNS, "datasette.io")
-        reply_queue = janus.Queue()
+        loop = asyncio.get_running_loop()
+        reply_future = loop.create_future()
         self._write_queue.put(
```
simonw, line 397: Note that even though we dropped the Janus queue we still queue writes using a standard library Queue, which is created here:

https://github.com/simonw/datasette/blob/46d90a0b887bfa23f986a28499eb5f85cd7eed04/datasette/database.py#L379-L383

simonw:
From the coverage report it looks like our tests don't exercise the various cases where an exception is raised during a query.

PR #2721, opened 2026-05-18
inspect: skip the table-count cap so on-disk numbers are exact

### LINKED ISSUE #2712: inspect-data.json caps row counts at 10,000
https://github.com/simonw/datasette/issues/2398 introduced code that caps row counts at 10,000 at run time. the inspect code shares this code path and now has row counts capped at 10,000. 

i don't know if this is something you want to fix given that datasette is moving away from the static use case.

c-tonneslan:
Closes #2712.

`Database.table_counts` has a `count_limit=10000` cap so the runtime UI doesn't sit on huge tables. The `datasette inspect` CLI shares that path, so the JSON it persists capped every row count at 10001.

This adds an `exact=True` path that issues a plain `SELECT COUNT(*)` (the same query `inspect_tables` in `inspect.py` already uses) and points the CLI at it. The exact path also skips caching the result on the `Database`, since a precomputed count is wrong if the file changes after inspect runs.

### DIFF
--- a/datasette/cli.py
+++ b/datasette/cli.py
@@ -154,7 +154,7 @@ async def inspect_(files, sqlite_extensions):
     app = Datasette([], immutables=files, sqlite_extensions=sqlite_extensions)
     data = {}
     for name, database in app.databases.items():
-        counts = await database.table_counts(limit=3600 * 1000)
+        counts = await database.table_counts(limit=3600 * 1000, exact=True)
         data[name] = {
             "hash": database.hash,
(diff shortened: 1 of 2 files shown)


simonw:
This adds a new `exact=True` parameter, but I'm going to solve this using this other route instead:

https://github.com/simonw/datasette/blob/74324cb8492be8aa8597e58fb6f690158128e6fc/datasette/inspect.py#L36-L50

PR #2722, opened 2026-05-18
actions_sql: always emit the anon_rules CTE so empty-anon path doesn't 500

### LINKED ISSUE #2644: --default-deny returns 500 on index pages when no config file is present
Running datasette --default-deny --root without a datasette.yaml config file causes a 500 Internal Server Error on the instance index (/) and database index pages.

## Steps to reproduce

In a fresh datasette install, with some test db, and **without a config file**.
```
datasette test.db --default-deny --root
```

Log in as root using the token link. Visit /.

### Expected

403 Forbidden or the normal index page (root is authenticated).

### Actual

Image

## Cause

Not sure about this cause, but generated by my model:

  In datasette/utils/actions_sql.py, the anon_rules CTE is conditionally defined (only when anon_sqls_rewritten is non-empty) but unconditionally referenced by anon_child_lvl, anon_parent_lvl, and anon_global_lvl.

  When there's no config file, default_action_permissions_sql(actor=None) returns None under --default-deny, all other hooks return empty for anonymous users, anon_sqls_rewritten stays empty, and the CTE is never created.

  A config file with allow: blocks avoids this because it generates deny rules for anonymous users, populating anon_sqls_rewritten.

#2593 assumes that the anon rules is always populated.

c-tonneslan:
Closes #2644.

`anon_rules` was only added when at least one anonymous permission rule came back from the hooks. With `--default-deny` and no config file the hooks all return nothing, but the downstream `anon_child_lvl` / `anon_parent_lvl` / `anon_global_lvl` CTEs `LEFT JOIN anon_rules` unconditionally, so the full query was invalid SQL and the index pages returned 500.

Emitting an empty CTE with the expected columns in the no-rules case keeps the LEFT JOINs happy without changing the result. Regression test under `test_default_deny_no_config_index_pages_dont_500`.

### DIFF
--- a/datasette/cli.py
+++ b/datasette/cli.py
@@ -154,7 +154,7 @@ async def inspect_(files, sqlite_extensions):
     app = Datasette([], immutables=files, sqlite_extensions=sqlite_extensions)
     data = {}
     for name, database in app.databases.items():
-        counts = await database.table_counts(limit=3600 * 1000)
+        counts = await database.table_counts(limit=3600 * 1000, exact=True)
         data[name] = {
             "hash": database.hash,
             "size": database.size,
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -561,25 +561,26 @@ class Database:
             self.cached_size = Path(self.path).stat().st_size
             return self.cached_size
 
-    async def table_counts(self, limit=10):
+    async def table_counts(self, limit=10, exact=False):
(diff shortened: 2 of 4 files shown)


simonw:
Looks like this also includes:
- https://github.com/simonw/datasette/pull/2721

I'll merge that one first.

PR #2726, opened 2026-05-21
fix: add keyboard accessibility and ARIA roles to cog menu

### LINKED ISSUE #1964: Cog menu is not keyboard accessible
This menu here: https://latest.datasette.io/fixtures/attraction_characteristic

You can tab to it (see the outline) and hit space to open it, but you can't then navigate the items in the open menu using the keyboard.

cog-menu

junliliu1:
The cog menu dropdown currently cannot be navigated by keyboard after opening — arrow keys don't move focus between items, and there are no ARIA attributes for screen readers.

This PR adds:
- `aria-haspopup="menu"` and `aria-expanded` on the `<summary>` toggle button, kept in sync via a `toggle` event listener
- `role="menu"` on the dropdown `<ul>` container
- `role="menuitem"` and `tabindex="-1"` on each menu item `<a>`
- Focus moves to the first menu item when the menu opens
- Arrow key navigation (Up/Down) between menu items
- Escape key closes the menu and returns focus to the toggle button

Changes are in `_action_menu.html` (ARIA markup) and `_close_open_menus.html` (keyboard event handling, which already wires up all `.details-menu` elements).

Fixes #1964

### DIFF
--- a/datasette/templates/_action_menu.html
+++ b/datasette/templates/_action_menu.html
@@ -1,7 +1,7 @@
 {% if action_links %}
 <div class="page-action-menu">
 <details class="actions-menu-links details-menu">
-    <summary>
+    <summary aria-haspopup="menu" aria-expanded="false">
         <div class="icon-text">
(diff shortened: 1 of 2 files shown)


junliliu1:
Closing in favor of #2727 submitted from the correct account.

PR #2727, opened 2026-05-21
fix: add keyboard accessibility and ARIA roles to cog menu

pintaste:
The cog menu dropdown currently cannot be navigated by keyboard after opening — arrow keys don't move focus between items, and there are no ARIA attributes for screen readers.

This PR adds:
- `aria-haspopup="menu"` and `aria-expanded` on the `<summary>` toggle button, kept in sync via a `toggle` event listener
- `role="menu"` on the dropdown `<ul>` container
- `role="menuitem"` and `tabindex="-1"` on each menu item `<a>`
- Focus moves to the first menu item when the menu opens
- Arrow key navigation (Up/Down) between menu items
- Escape key closes the menu and returns focus to the toggle button

### DIFF
--- a/datasette/templates/_action_menu.html
+++ b/datasette/templates/_action_menu.html
@@ -1,7 +1,7 @@
 {% if action_links %}
 <div class="page-action-menu">
 <details class="actions-menu-links details-menu">
-    <summary>
+    <summary aria-haspopup="menu" aria-expanded="false">
         <div class="icon-text">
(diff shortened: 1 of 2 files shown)


simonw:
I just tried VoiceOver on my Mac (Command+F5) and confirmed the column cog menus cannot be tabbed to:

tab

simonw:
My test was actually against a different cog menu though.

simonw:
I tested this further here and it works well:
- #2738 

See demo of alternatives here: https://static.simonwillison.net/static/2026/action-menu-accessibility.html
 
Landing as-is, thanks!

pintaste:
Thanks for the thoughtful review and for taking the time to validate the VoiceOver behavior carefully. Happy to contribute. Looking forward to improving accessibility further in Datasette.

PR #2732, opened 2026-05-22
/-/jump, plus jump_items_sql() and makeJumpSections() plugin hooks

### LINKED ISSUE #2731: Turn "/" jump to table menu into plugin expandable jump to X
I really like the "Jump to" table menu that's available via `/` and now, as-of #2725, available at the top of the application menu.

Currently it only works to navigate to tables. I want it to be able to also navigate to:

- databases
- canned queries
- other stuff added by plugins

It's also a useful piece of real-estate in the UI - the blank slate before you start typing is currently empty, but maybe plugins should be able to populate that with useful things?

simonw:
Refs:
- #2731 

Codex transcript for this prototype: https://gist.github.com/simonw/c0dafded42c4accd7b9d6985f621ad95

----
📚 Documentation preview 📚: https://datasette--2732.org.readthedocs.build/en/2732/


datasette/static/navigation-search.js
```diff
@@ -306,79 +362,253 @@ class NavigationSearch extends HTMLElement {
...
+        description: match.description ? String(match.description) : "",
+      };
+      const recentItems = this.loadRecentItems().filter(
+        (recentItem) => recentItem.url !== item.url,
+      );
+      localStorage.setItem(
+        this.recentItemsStorageKey(),
+        JSON.stringify([item, ...recentItems].slice(0, 5)),
+      );
+    } catch (e) {
```
simonw, line 453: I added a mechanism to show the most recently used items in the jump to menu empty state.

datasette/templates/base.html
```diff
@@ -70,7 +70,7 @@
 {% endfor %}
 
 {% if select_templates %}<!-- Templates considered: {{ select_templates|join(", ") }} -->{% endif %}
-<script src="{{ urls.static('navigation-search.js') }}" defer></script>
-<navigation-search url="/-/tables"></navigation-search>
+<script src="{{ urls.static('navigation-search.js') }}{% if navigation_search_js_hash is defined %}?{{ navigation_search_js_hash }}{% endif %}" defer></script>
```
simonw: Ditch the `navigation_search_js_hash` stuff.

datasette/app.py
```diff
@@ -1219,13 +1219,24 @@ def _plugin_config_nested(self, plugin_name, database, table=None, fallback=True
...
+        return static_hash
+
     def app_css_hash(self):
-        if not hasattr(self, "_app_css_hash"):
-            with open(os.path.join(str(app_root), "datasette/static/app.css")) as fp:
-                self._app_css_hash = hashlib.sha1(fp.read().encode("utf8")).hexdigest()[
-                    :6
-                ]
-        return self._app_css_hash
+        return self.static_hash("app.css")
```
simonw, line 1239: This refactor snuck in for the feature I removed in 9e7419db8deb42c6236b6d6a142f6b225e010573 but I like it so I'm leaving it.

datasette/default_jump_items.py
```diff
@@ -0,0 +1,119 @@
+from datasette import hookimpl
+from datasette.jump import JumpSQL
+
+
+async def _query_display_names_sql(datasette, actor):
+    selects = []
+    params = {}
+    for database_name in datasette.databases.keys():
+        queries = await datasette.get_canned_queries(database_name, actor)
```
simonw: I'm not crazy about this. We loop through each database, fetch all of its canned queries and then use them to construct a huge UNION query to represent the very canned query - with named SQL parameters to insert the correct values for them.

`MAX_VARIABLE_NUMBER` in SQLite uses to be 999 but is higher now. Still worth keeping an eye on.

simonw: The solution to this will be to completely rethink how canned queries work - they need to move into the internal database, not just for this but for other things I want to do with them too.

Filed an issue:
- https://github.com/simonw/datasette/issues/2735

datasette/default_jump_items.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        table_sql, table_params = await datasette.allowed_resources_sql(
+            action="view-table", actor=actor
+        )
+        query_sql, query_params = await datasette.allowed_resources_sql(
+            action="view-query", actor=actor
+        )
+        query_display_names_sql, query_display_names_params = (
+            await _query_display_names_sql(datasette, actor)
+        )
```
simonw: The only reason we use both the `allowed_resources_sql()` thing and the `_query_display_names_sql()` thing is so we can show display names for the canned queries that might be in the config file.

Let's simplify and drop the display names for the moment, we can bring them back in:
- #2735

datasette/default_jump_items.py
simonw: Removed this issue in 1590444fa34c132eaecf49391033925b8fb123a8

docs/plugin_hooks.rst
```diff
@@ -1881,6 +1881,84 @@ Using :ref:`internals_datasette_urls` here ensures that links in the menu will t
...
+The SQL query must return these columns:
+
+``type``
+    A short type string for the result, for example ``"app"`` or ``"dashboard"``.
+
+``label``
+    The stable name for the result. This is returned as ``name`` in the JSON API and is used for sorting.
+
+``description``
+    A short description shown above the item in the jump menu.
```
simonw: The `description` here is actually something like "Database" or "Canned query" - it's a human-readable version of the `type`, not a description of the item itself. This is confusing and feels unnecessary.

simonw:
I'm going to use `description` in the SQL query as an optional per-item description which is displayed truncated like this instead:

CleanShot 2026-05-23 at 17 13 14@2x

simonw:
https://github.com/simonw/datasette/blob/9c1f8621eb9f339fc0ba59bb6e6ca2cb77942fb0/docs/plugin_hooks.rst?plain=1#L1913-L1920

The SQL queries were returning `database_name` and `resource_name` here purely so Datasette could use `datasette.urls.table(...)` and friends to construct the correct URL, because a SQL query can't correctly escape the table/database names itself.

I'm changing that to just returning a `url` column but optionally supporting JSON in that like so:

```sql
select json_object(
  'method', 'table',
  'database', allowed_tables.parent,
  'table', allowed_tables.child
) AS url
```

simonw:
Here's the documentation for what I think is the final design of that `jump_items_sql` hook:

https://github.com/simonw/datasette/blob/c73ed1ee4e619ee9da002443da3e29e0bc93de4f/docs/plugin_hooks.rst#jump_items_sqldatasette-actor-request

simonw:
I'm going to drop the `sort_key` and `source` columns, to further simplify the design.

simonw:
One last thing to fix: right now `JumpSQL` only works against data in the internal database, but future plugins may well want to include items in the `/-/jump` menu from other databases as well.

I'm going to add a `database=` optional parameter to `JumpSQL` and have the `/-/jump` page combine the results of queries against multiple databases in Python code.

docs/plugin_hooks.rst
simonw: Fixed that, it's now shown as a `...` CSS truncated description under the item, if set.

simonw:
This feels really good:

slash-menu

Issue #2735, opened 2026-05-23
Rebrand "canned queries" to "queries", put them in internal database


simonw:
Currently canned queries live in configuration or can be provided by the `canned_queries` plugin hook:

https://github.com/simonw/datasette/blob/1000d50220faef339ee688aae9150e83a205be3a/datasette/hookspecs.py#L140-L142

This does not scale well to hundreds or thousands of canned queries. I want to move them into the internal database, similar to what we did with metadata.

This likely means retiring the existing plugin hook in favor of `await datasette.add_canned_query(...)` / `.remove_canned_query()` style methods that plugins can use.

simonw:
This just came up in the `/-/jump` work, since I'd like to autocomplete against names of canned queries so they need to be efficiently available, ideally in that internal database.

eyeseast:
This would definitely help a plugin I use in every project: https://github.com/eyeseast/datasette-query-files

simonw:
I made a bunch of progress on the design for this in some prototyping. Here's how it's shaping up:
- `queries` table in the internal database to store these
- `datasette.add_query()`, `datasette.update_query()`, `datasette.remove_query()`, `datasette.get_query()` asyncio Python API
- Several new permissions:
    - `insert-query` - user can save a query, it belongs to them
    - `publish-query` - user can "publish" a query so people other than them can see it
    - `update-query` - for updating queries not owned by the current user
    - `delete-query` - for deleting queries not owned by the current user
- UI for saving queries

Then things got more complicated as I started thinking about writable queries:

- Need a way to check that the user has permission to insert/update data into all tables that a writable query they are composing needs to use. Figured out a mechanism for this with SQLite3 authorizers to analyze a query.
- Doesn't make sense for a user to be able to save and then execute a named writable query if they can't also execute arbitrary write queries... so I added a `execute-write-sql` permission and accompanying UI, available via a database action menu. This still checks the required table permissions before allowing the execute though.

simonw:
The permissions model is a bit tricky to figure out, because a user sharing a query with other users might allow those other users to access data (or update data) that they would otherwise not be able to access.

Being able to enforce permissions based on analyzing the query to see what tables it interacts with might turn out to be a good way to resolve this.

Still need to think about exposing queries to the general public though, especially from private instances. Starting those with the `datasette   --default-deny` option may be the right way to think about this, since then individual queries need to have `view-query` permission opened up on them.

simonw:
[...]

PR #2741, opened 2026-05-25
Write SQL query UI, canned queries (including writable) now stored in internal database

simonw:
Refs:
- #2735 

----
📚 Documentation preview 📚: https://datasette--2741.org.readthedocs.build/en/2741/


datasette/default_permissions/defaults.py
```diff
@@ -21,7 +21,6 @@
         "view-database",
         "view-database-download",
         "view-table",
-        "view-query",
```
simonw, line 24: This looks like a mistake.

datasette/default_permissions/__init__.py
```diff
@@ -35,14 +28,6 @@
 from .defaults import (
     default_allow_sql_check as default_allow_sql_check,
     default_action_permissions_sql as default_action_permissions_sql,
+    default_query_permissions_sql as default_query_permissions_sql,
     DEFAULT_ALLOW_ACTIONS as DEFAULT_ALLOW_ACTIONS,
 )
```
simonw, line 34: Why is this using `x as x`?

datasette/default_permissions/defaults.py
```diff
@@ -21,12 +21,37 @@
...
 
+def _configured_query_restriction_selects(datasette: "Datasette") -> tuple[list[str], dict]:
+    selects = []
+    params = {}
+    for index, (database_name, db_config) in enumerate(
+        ((datasette.config or {}).get("databases") or {}).items()
+    ):
+        for query_name, query_config in (db_config.get("queries") or {}).items():
+            if isinstance(query_config, dict) and query_config.get("is_private"):
+                continue
```
simonw: I don't think we need to look at queries in config here - they should have been loaded elsewhere.

datasette/default_permissions/defaults.py
```diff
@@ -67,3 +67,52 @@ async def default_action_permissions_sql(
...
+
+@hookimpl(specname="permission_resources_sql")
+async def default_query_permissions_sql(
+    datasette: "Datasette",
+    actor: Optional[dict],
+    action: str,
+) -> Optional[PermissionSQL]:
+    actor_id = actor.get("id") if isinstance(actor, dict) else None
+
+    if action in {"update-query", "delete-query"}:
```
simonw: Need to be sure that anyone granted update-query/delete-query globally still cannot delete or update private queries that do not belong to them.

simonw:
Improved query facet UI:

CleanShot 2026-05-26 at 13 06 35@2x

simonw:
Create saved query UI now looks like this:

CleanShot 2026-05-26 at 13 47 55@2x

datasette/default_permissions/__init__.py
simonw: Turns out this pattern is there to avoid a `mypy` error:

> `error: Module "datasette.default_permissions" does not explicitly export attribute`

simonw:
I'm going to rename the `insert-query` permission - which is a little confusing since "insert" is a SQL keyword - to `store-query` instead. I'll keep `update-query` and `delete-query` though.

datasette/hookspecs.py
```diff
@@ -234,7 +229,7 @@ def top_query(datasette, request, database, sql):
 
 @hookspec
 def top_canned_query(datasette, request, database, query_name):
```
simonw: Rename this to `top_query`.

.github/workflows/deploy-latest.yml
```diff
@@ -69,23 +69,24 @@ jobs:
...
-                        "on_success_message_sql": "select 'Counter {name} decremented to ' || value from counters where name = '{name}'".format(name=name),
-                        "write": True,
-                    }
-                return queries
+                    await datasette.add_query(
+                        "counters",
+                        "increment_{}".format(name),
+                        "update counters set value = value + 1 where name = '{}'".format(name),
+                        on_success_message_sql="select 'Counter {name} incremented to ' || value from counters where name = '{name}'".format(name=name),
+                        is_write=True,
```
simonw, line 78: This previously used the `canned_queries()` hook, now removed - so now it uses the startup hook instead.

simonw:
I had Codex and GPT-5.5 xhigh run a review of the branch against main in a fresh session:

> Run git diff against main, do a thorough review of this branch - we are about to land it, need to make sure we have not missed anything important

It found three problems that absolutely need to be resolved: https://gist.github.com/simonw/d59a02c29c816cd46d5f9018252ab483

----

**Findings**
- [P1] Trusted queries can be edited and stay trusted. Config queries default to trusted in [datasette/app.py](/Users/simon/Dropbox/dev/datasette/datasette/app.py:622), and trusted execution skips SQL/write checks in [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:700). But `QueryUpdateView` lets anyone with `update-query` change the query SQL and preserves `is_trusted`, see [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:1557). I confirmed a viewer with only `view-query` could run an edited trusted query and read data they otherwise could not. I would block updates to trusted/config queries, or clear `is_trusted` on any API/user update to SQL-bearing fields.

- [P1] User/API stored queries can persist raw HTML/JS. `description_html` is accepted in stored query create/update fields at [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:434), passed through on create at [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:883), then rendered through the existing `|safe` metadata path in [_description_source_license.html](/Users/simon/Dropbox/dev/datasette/datasette/templates/_description_source_license.html:4). I reproduced a stored query page rendering `<script>window.XSS=1</script>`. User-created/API-updated queries should reject `description_html`, sanitize it, or only allow it for config/trusted sources.

- [P2] `on_success_message_sql` can be injected via update without `execute-sql`. The update view only checks `execute-sql` when the `"sql"` field changes, at [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:1576), but `_prepare_query_update()` also accepts `on_success_message_sql` at [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:942). That SQL later executes after POST at [database.py](/Users/simon/Dropbox/dev/datasette/datasette/views/database.py:1693). I confirmed an actor with `update-query` only could make a trusted write query return a secret as its success message.

simonw:
I have reviewed all of the code! Still to do:

- [x] The three issues identified above: https://github.com/simonw/datasette/pull/2741#issuecomment-4549620486
- [x] Documentation for the `datasette.add_query()` etc methods, since those replace the `canned_queries()` plugin hook that we removed.

simonw:
For all three of the cases in https://github.com/simonw/datasette/pull/2741#issuecomment-4549620486 I think the easiest solution for the moment is to disallow setting `is_trusted`, `on_success_message_sql` or `description_html` from the UI or API - only allow setting those in configuration.

datasette/stored_queries.py
```diff
@@ -0,0 +1,541 @@
...
+        "is_private": bool(row["is_private"]),
+        "is_trusted": bool(row["is_trusted"]),
+        "source": row["source"],
+        "owner_id": row["owner_id"],
+        "on_success_message": options.get("on_success_message"),
+        "on_success_message_sql": options.get("on_success_message_sql"),
+        "on_success_redirect": options.get("on_success_redirect"),
+        "on_error_message": options.get("on_error_message"),
+        "on_error_redirect": options.get("on_error_redirect"),
+    }
```
simonw: Would be neater if this was a dataclass not a dict.

simonw: Done in https://github.com/simonw/datasette/pull/2741/commits/b1289a73f9869e83a433a088c2a6c48285e67f2d

Issue #2743, opened 2026-05-25
latest.datasette.io/extra_database/-/query 500 error


simonw:
https://latest.datasette.io/extra_database/-/query without any parameters currently returns a 500 error:

> expected string or bytes-like object, got 'NoneType'

simonw:
Deployed.

https://latest.datasette.io/extra_database/-/query no longer returns an error.

https://latest.datasette.io/extra_database/-/query.csv returns a 400 error "?sql= is required" which is fair enough.

PR #2744, opened 2026-05-25
Fix 500 error when accessing query page without ?sql= parameter

simonw:
> Use `curl` to visit https://latest.datasette.io/extra_database/-/query - then investigate the error

Refs:
- #2743

https://claude.ai/code/session_013vfRb8GU5itMZ9PCR95hYa

### DIFF
--- a/datasette/templates/query.html
+++ b/datasette/templates/query.html
@@ -46,14 +46,14 @@
     {% if not hide_sql %}
         {% if editable and allow_execute_sql %}
             <p><textarea id="sql-editor" name="sql"{% if query and query.sql %} style="height: {{ query.sql.split("\n")|length + 2 }}em"{% endif %}
-            >{% if query and query.sql %}{{ query.sql }}{% else %}select * from {{ tables[0].name|escape_sqlite }}{% endif %}</textarea></p>
+            >{% if query and query.sql %}{{ query.sql }}{% elif tables %}select * from {{ tables[0].name|escape_sqlite }}{% endif %}</textarea></p>
         {% else %}
             <pre id="sql-query">{% if query %}{{ query.sql }}{% endif %}</pre>
         {% endif %}
     {% else %}
         {% if not canned_query %}
             <input type="hidden" name="sql"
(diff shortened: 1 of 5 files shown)


simonw:
Claude Code said:

----

## Summary
Fixed a bug where visiting `/<database>/-/query` without a `?sql=` parameter would return a 500 error instead of rendering the empty SQL editor form.

## Key Changes
- Added null checks for `sql` variable before attempting to derive named parameters, validate SQL, or process query execution
- Updated the CSV format handler to return a 400 error with a clear message when `?sql=` is missing
- Modified the query template to only show default SQL suggestions when tables are available
- Added guards in the plugin hook to handle cases where SQL is not provided
- Added test cases to verify the query page renders correctly without SQL and that CSV requests without SQL return 400

## Implementation Details
- The main issue was that the code assumed `sql` would always be present when processing database queries
- Added conditional checks (`if sql:`) before operations that depend on SQL being defined:
  - Named parameter derivation
  - SQL validation for edit URL generation
  - CSV data fetching
- The query template now uses `elif tables` instead of `else` to avoid showing placeholder SQL when no tables are available
- Plugin hooks are now defensive against missing SQL parameters


----
📚 Documentation preview 📚: https://datasette--2744.org.readthedocs.build/en/2744/

datasette/templates/query.html
```diff
@@ -46,14 +46,14 @@ <h3>Custom SQL query{% if display_rows %} returning {% if truncated %}more than
     {% if not hide_sql %}
         {% if editable and allow_execute_sql %}
             <p><textarea id="sql-editor" name="sql"{% if query and query.sql %} style="height: {{ query.sql.split("\n")|length + 2 }}em"{% endif %}
-            >{% if query and query.sql %}{{ query.sql }}{% else %}select * from {{ tables[0].name|escape_sqlite }}{% endif %}</textarea></p>
+            >{% if query and query.sql %}{{ query.sql }}{% elif tables %}select * from {{ tables[0].name|escape_sqlite }}{% endif %}</textarea></p>
```
simonw, line 49: It spotted and fixed this bug too, which occurred in databases with 0 tables.

PR #2745, opened 2026-05-26
Add content hash to JS includes (closes #2714)

### LINKED ISSUE #2714: Datasette JS should include content hash in the URL
datasette css already includes a content hash #154. it would be good if the included js did this same.

mvanhorn:
## Summary

Adds `?{{ static_hash('foo.js') }}` to the five first-party JS includes shipped with datasette so browsers refetch when the bundled file changes, matching the existing pattern used for `app.css`.

Vendored libraries (`cm-editor-6.0.1.bundle.js`, `sql-formatter-2.3.3.min.js`, `json-format-highlight-1.0.1.js`) already carry a version in the filename and were left alone.

Closes #2714

## Approach

`Datasette.static_hash(filename)` already exists (it's what `app_css_hash` calls under the hood). The change exposes it as a callable in the template context so templates can reuse the same SHA-1-prefix scheme as `app.css` without hardcoding per-file context entries.

`handle_exception.py` also gets `static_hash` added to the error-page template context, the same way it currently passes `app_css_hash`.

## Files changed

- `datasette/app.py` — expose `static_hash` callable.
- `datasette/handle_exception.py` — include it in the error template context.
- `datasette/templates/base.html` — hash `datasette-manager.js` and `navigation-search.js`.
- `datasette/templates/table.html` — hash `column-chooser.js`, `table.js`, `mobile-column-actions.js`.

## Tests

- New `test_js_content_hash` parametrized across all 5 files, asserts the rendered `<script src>` includes the expected `?hash` suffix.
- Existing `test_navigation_menu_links` updated to expect the new query string on `navigation-search.js`.
- Full `tests/test_html.py` + `tests/test_table_html.py` pass locally (212 passed, 1 xpassed).

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -2052,6 +2052,7 @@ class Datasette:
                 and "ds_actor" in request.cookies
                 and request.actor,
                 "app_css_hash": self.app_css_hash(),
+                "static_hash": self.static_hash,
                 "zip": zip,
                 "body_scripts": body_scripts,
                 "format_bytes": format_bytes,
--- a/datasette/handle_exception.py
+++ b/datasette/handle_exception.py
@@ -67,6 +67,7 @@ def handle_exception(datasette, request, exception):
                         info,
                         urls=datasette.urls,
                         app_css_hash=datasette.app_css_hash(),
+                        static_hash=datasette.static_hash,
                         menu_links=lambda: [],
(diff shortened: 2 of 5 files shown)


simonw:
I handled this a different way:
- #2804

PR #2755, opened 2026-05-29
Return MultiParams from request.post_vars() to handle multi-valued keys

### LINKED ISSUE #2425: `request.post_vars()` cannot handle multiple values for a key
https://github.com/simonw/datasette/blob/832f76ce26ffb2f3e27a006ff90254374bd90e61/datasette/utils/asgi.py#L138-L140

Datasette GET requests often use `?_facet=category&_facet=size` - but right now there's no neat way to handle that with POST data, since this code reduces any multiple values to just one.

This came up while working with `<select multiple>` (enhanced using Choices.js) in:
- https://github.com/datasette/datasette-acl/issues/24

aicayzer:
Closes #2425.

## Problem

``request.post_vars()`` is implemented as ``dict(parse_qsl(body, keep_blank_values=True))``, which silently collapses any duplicate keys to the last value. Forms that submit multiple values for the same field — e.g. ``<select multiple>``, multi-checkbox groups, ``?_facet=...&_facet=...`` style POSTs — lose data with no error. ``request.args`` already handles this correctly via ``MultiParams``.

## Fix

Return a ``MultiParams`` built from ``parse_qs(...)`` instead of a ``dict``, matching the shape of ``request.args``. ``MultiParams`` was already imported in ``asgi.py``, so no new imports.

Added an ``items()`` method to ``MultiParams`` that yields ``(key, first_value)`` pairs (matching ``__getitem__`` semantics), so ``MultiParams`` can be passed directly into ``dict()``, dict-comprehension patterns and similar.

The internal ``_json_or_form_payload`` helper feeds two callers — ``_coerce_execute_write_payload`` and the ``QueryStoreView`` / ``QueryUpdateView`` POST handlers — that both had ``isinstance(data, dict)`` guards intended to validate JSON shape. Those guards spuriously rejected the new ``MultiParams`` on the form path, so they have been gated on ``is_json``. The fall-through logic that re-renders the HTML form on validation errors uses ``.get()`` on ``query_data`` and works against either ``dict`` or ``MultiParams``.

## Behavioural notes

- ``post_vars[key]`` and ``post_vars.get(key)`` continue to work but now return the *first* submitted value rather than the *last*. This matches ``request.args`` semantics and is documented in the changelog.
- ``post_vars.getlist(key)`` now returns the full list of submitted values — the new capability the issue asked for.
- Plugins doing ``Response.json(await request.post_vars())`` should wrap in ``dict(...)`` because ``MultiParams`` is not directly JSON-serialisable. The bundled ``my_plugin`` test fixture has been updated to demonstrate this.

## Test plan

- Updated ``test_request_post_vars`` to assert ``MultiParams`` behaviour against the previously-existing single-value payload.
- New ``test_request_post_vars_multi`` mirrors ``test_request_args``, covering ``[]``, ``get``, ``getlist``, ``in``, ``keys``, ``len`` and ``KeyError`` for missing keys.
- Full test suite passes (``pytest --ignore=tests/test_multipart.py -n auto``: 1853 passed). ``tests/test_multipart.py`` has a pre-existing ``ModuleNotFoundError: multipart_form_data_conformance`` unrelated to this change.

## Files touched

- ``datasette/utils/asgi.py`` — ``post_vars()`` returns ``MultiParams``.
- ``datasette/utils/__init__.py`` — ``MultiParams.items()``.
- ``datasette/views/query_helpers.py`` — gate ``isinstance(data, dict)`` on ``is_json`` in ``_coerce_execute_write_payload``.
- ``datasette/views/stored_queries.py`` — gate ``isinstance`` guards on ``is_json`` in ``QueryStoreView.post`` and ``QueryUpdateView.post``; drop the now-redundant ``isinstance(query_data, dict)`` checks in the error fall-through branches.
- ``tests/test_internals_request.py`` — updated and added tests.
- ``tests/plugins/my_plugin.py`` — wrap ``post_vars()`` in ``dict()`` for JSON serialisation.
- ``docs/internals.rst`` and ``docs/changelog.rst`` — documentation updates.

### DIFF
--- a/datasette/utils/__init__.py
+++ b/datasette/utils/__init__.py
@@ -1193,6 +1193,11 @@ class MultiParams:
         """Return full list"""
         return self._data.get(name) or []
 
+    def items(self):
+        """Yield (key, first_value) pairs, matching ``__getitem__`` semantics."""
+        for key, values in self._data.items():
+            yield key, values[0]
+
 
 class ConnectionProblem(Exception):
     pass
--- a/datasette/utils/asgi.py
+++ b/datasette/utils/asgi.py
@@ -3,7 +3,7 @@ import re
 from http.cookies import Morsel, SimpleCookie
 from mimetypes import guess_type
 from pathlib import Path
-from urllib.parse import parse_qs, parse_qsl, urlunparse
+from urllib.parse import parse_qs, urlunparse
 
 import aiofiles
 import aiofiles.os
@@ -194,7 +194,7 @@ class Request:
 
     async def post_vars(self):
         body = await self.post_body()
-        return dict(parse_qsl(body.decode("utf-8"), keep_blank_values=True))
+        return MultiParams(parse_qs(qs=body.decode("utf-8"), keep_blank_values=True))
 
     async def json(self):
         body = await self.post_body()
--- a/datasette/views/query_helpers.py
+++ b/datasette/views/query_helpers.py
@@ -281,9 +281,9 @@ def _execute_write_disabled_reason(sql, analysis_error, analysis_rows):
 
 
 def _coerce_execute_write_payload(data, is_json):
-    if not isinstance(data, dict):
-        raise QueryValidationError("JSON must be a dictionary")
     if is_json:
+        if not isinstance(data, dict):
+            raise QueryValidationError("JSON must be a dictionary")
         invalid_keys = set(data) - {"sql", "params"}
         if invalid_keys:
             raise QueryValidationError(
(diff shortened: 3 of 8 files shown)


aicayzer:
rebased onto main to clear the conflict that had crept in. also spotted the new query edit form handler had the same dict-shape check that needed gating on is_json for the multiparams return, so folded that in. full suite green locally, ready when you want a look.

aicayzer:
PR has been open for too long; closing to avoid staleness.

PR #2758, opened 2026-05-30
Fix /-/check 500 error on query-scoped actions

### LINKED ISSUE #2756: QueryResource.__init__() got an unexpected keyword argument 'table'
I got that on this page (signed in as root):

https://latest.datasette.io/-/check?action=delete-query

Image

truffle-dev:
Closes #2756.

`_check_permission_for_actor` builds child-level resources with hardcoded `table=child`, which raises `TypeError` for `QueryResource` (it accepts `query=`, not `table=`). Visiting `/-/check?action=delete-query` (or `view-query`, `update-query`) returned a 500.

Switched the call to positional args so it works with every `Resource` subclass regardless of the second-parameter name.

Verification: with the fix reverted via `git stash`, the three new parametrized cases reproduce the original `TypeError`; restoring it makes them pass. Full `tests/test_permissions.py` and `tests/test_actions_sql.py` runs are green.

### DIFF
--- a/datasette/views/special.py
+++ b/datasette/views/special.py
@@ -497,11 +497,13 @@ async def _check_permission_for_actor(ds, action, parent, child, actor):
     if action_obj.resource_class is None:
         resource_obj = None
     elif action_obj.takes_parent and action_obj.takes_child:
-        # Child-level resource (e.g., TableResource, QueryResource)
-        resource_obj = action_obj.resource_class(database=parent, table=child)
+        # Child-level resource (e.g., TableResource, QueryResource). Pass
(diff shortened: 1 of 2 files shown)


truffle-dev:
Closing as redundant — this is already fixed on `main` by d5141a57 ("Fix /-/check 500 for query actions (#2756)", shipped in 1.0a33). That commit lands the same fix this PR proposed (passing the resource child positionally so `QueryResource` and other subclasses with a differently-named second parameter construct correctly) plus an equivalent test, so there's nothing left here to merge. Thanks for sorting it.

PR #2763, opened 2026-05-31
Fixes for SQL write with RETURNING

### LINKED ISSUE #2762: SQL write with RETURNING: cannot commit transaction - SQL statements in progress
I tried a stored query like this:

    update table set name = :name where id = :id returning id

And got an error:

    cannot commit transaction - SQL statements in progress

Turns out the way `execute_write()` is implemented in Datasette causes the commit to end while the cursor full of returned rows has not yet been consumed, which results in an error.

simonw:
Closes:
- #2762 

----
📚 Documentation preview 📚: https://datasette--2763.org.readthedocs.build/en/2763/

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -2870,19 +2870,22 @@ def wrap_view_function(view_fn, datasette):
 
 
 def permanent_redirect(path, forward_query_string=False, forward_rest=False):
-    return wrap_view(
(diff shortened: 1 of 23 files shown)


simonw:
Tests are failing because SQLite 3.25.0 (the earliest version Datasette supports) does not support RETURNING.

Issue #2761, opened 2026-05-31
Cannot execute a write query with a parameter called :sql


simonw:
<img width="1438" height="1272" alt="Image" src="https://github.com/user-attachments/assets/0a864939-c9f3-4c46-9d8c-2fbf95117ac8" />

I tried doing this, but the `sql` parameter I entered over-rode the `sql` parameter for the rest of the form!

PR #2765, opened 2026-06-01
Fix write query failing when a named parameter is called :sql

viraatdas:
Fixes:
- #2761

When you run a write query containing a named parameter called `:sql` (for example `update docs set title = :sql where id = :id`), the generated parameter input on the execute-write form is named `sql`, which collides with the form's own `sql` field holding the query text. The user's `:sql` value then overrides the query for the rest of the form, so the write does not work.

This namespaces the generated parameter form inputs with a `_sql_param_` prefix (via a new `sql_parameter_name_prefix`), so a `:sql` (or `:id`, etc.) parameter input no longer collides with the control `sql` field. The label still shows the bare parameter name, and the JS parameter controls track the original name via `data-parameter-name`; the execute-write view reads the prefixed fields.

Added an end-to-end regression test (`test_execute_write_form_parameter_called_sql`) that runs a write query with a `:sql` parameter and asserts the affected row is updated correctly. It fails before this change.

Prepared with AI assistance.

### DIFF
--- a/datasette/templates/_sql_parameter_scripts.html
+++ b/datasette/templates/_sql_parameter_scripts.html
@@ -22,16 +22,15 @@ window.datasetteSqlParameters = (() => {
     };
   }
 
-  function parameterName(control) {
-    return control.dataset.parameterName || control.name;
-  }
-
   function syncParameterState(manager) {
     manager.parameterState = new Map();
     manager.section
       .querySelectorAll("[data-parameter-control]")
       .forEach((control) => {
-        manager.parameterState.set(parameterName(control), controlState(control));
+        manager.parameterState.set(
+          control.dataset.parameterName,
+          controlState(control)
+        );
       });
   }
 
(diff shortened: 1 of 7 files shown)


simonw:
Here's how I'm manually testing this:
```bash
echo '{"id": 1, "title": "Initial"}' | sqlite-utils insert /tmp/demo.db docs --pk id - 
uv run datasette /tmp/demo.db --root
```
Then on http://127.0.0.1:8001/demo/-/execute-write this SQL:

```sql
update docs set title = :sql where id = :id
```

The two form fields appear:

CleanShot 2026-06-10 at 20 11 34@2x

Submitting the form works:

CleanShot 2026-06-10 at 20 11 58@2x

There's a server-side flow too:

http://127.0.0.1:8001/demo/-/execute-write?sql=update+docs+set+title+=+:sql+where+id+=+:id

That one works the same way.

Without this patch I get this error:

CleanShot 2026-06-10 at 20 14 25@2x

PR #2773, opened 2026-06-12
Clearer examples and descriptions for JSON API extras

simonw:
> Review the most recent addition to the JSON API documentation
> 
> Some of the extras do not have clear examples of their output, and some of the descriptions may not be clear enough. Fix that, including linking to relevant other sections of documentation

Review of the generated ?_extra= documentation found several extras
with no example output or with examples that needed explanation:

- extras: now shows an abbreviated example of the toggle list and has
  a clearer description (which also improves the live API output)
- set_column_type_ui: example of the shape seen with set-column-type
  permission, plus a note that it is null otherwise
- column_types: live example generated from a table with an assigned
  column type instead of an empty {}
- metadata: live table example now demonstrates a table description
  and column descriptions; row and query examples gained explanatory
  notes
- expandable_columns, foreign_key_tables, facets_timed_out, next_url,
  renderers: notes explaining the shape of their output

Also added docs_note cross-references to the relevant documentation:
facets, pagination, render_cell and register_output_renderer plugin
hooks, column type configuration and API, metadata, custom templates,
permissions and foreign key label expansion. foreign_key_tables is
now flagged as potentially executing additional queries.

https://claude.ai/code/session_01EfjBe6E817m9XNFW7EX3Vm

----
📚 Documentation preview 📚: https://datasette--2773.org.readthedocs.build/en/2773/

### DIFF
--- a/datasette/views/table_extras.py
+++ b/datasette/views/table_extras.py
@@ -184,6 +184,7 @@ class FacetResultsExtra(Extra):
     )
     scopes = {ExtraScope.TABLE}
     expensive = True
+    docs_note = "See :ref:`facets` for details of how facets work."
 
     async def resolve(self, context, facet_instances):
         facet_results = {}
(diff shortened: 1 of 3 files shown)


simonw:
Reviewed here: https://datasette--2773.org.readthedocs.build/en/2773/json_api.html?readthedocs-diff=true&readthedocs-diff-chunk=1#table-json-responses

PR #2777, opened 2026-06-14
fix: remove chunked database download header

### LINKED ISSUE #2663: Database download doesn't show progress bar in browsers
It sends a content length header but browsers don't show a progress bar for the download.

I think that's because of this workaround for Cloud Run from a few years ago:

- #749

suzunn:
## Problem

Database downloads include both a `Content-Length` header and an explicit `Transfer-Encoding: chunked` header. Browsers do not show download progress for that response shape, even though the file size is already known.

Closes #2663.

## Solution

I removed the manual `Transfer-Encoding: chunked` header from immutable database downloads and updated the existing download test to assert that the response keeps `Content-Length` without setting `transfer-encoding`.

## Testing

- `uv run ruff check datasette\views\database.py tests\test_html.py`
- `uv run black --check datasette\views\database.py tests\test_html.py`
- `git diff --check`

I also ran `uv run pytest tests\test_html.py -k database_download_for_immutable -q`. The response assertions completed, but the test fails during Windows temporary directory cleanup with `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`; that looks like the existing Windows file-handle cleanup issue tracked separately from this header change.

### DIFF
--- a/datasette/views/database.py
+++ b/datasette/views/database.py
@@ -444,7 +444,6 @@ async def database_download(request, datasette):
         if_none_match = request.headers.get("if-none-match")
         if if_none_match and if_none_match == etag:
             return Response("", status=304)
-    headers["Transfer-Encoding"] = "chunked"
     return AsgiFileDownload(
         filepath,
         filename=os.path.basename(filepath),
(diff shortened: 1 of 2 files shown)


suzunn:
Friendly ping on this small fix whenever you have a moment. It's conflict-free and CI is green. Happy to add a test or adjust the approach if that would help it along.

PR #2781, opened 2026-06-14
Insert/edit/delete UI in Datasette core, plus makeColumnField() plugin hook

### LINKED ISSUE #2780: Insert and edit UI as part of Datasette core
We've previously handled this in the https://github.com/datasette/datasette-write-ui plugin but it really belongs in Datasette core, for a few reasons:

1. We have a JSON write API in core now, and it's always a bit weird when there's a feature that's only an API but has no UI
2. We added custom column types in #2664
    - So now editing needs to be column-type aware... but we also need some kind of JavaScript plugin hook so custom column types from plugins can customize the form. Any time a plugin needs extra plugin hooks is a good point to consider moving it to core.
3. As Datasette Agent grew the ability to edit data in the database it suddenly became clear that NOT having a non-agent UI for doing that was confusing and inconsistent

simonw:
Closes:
- #2780

There's a lot of stuff in this branch:

- A new `/db/table/-/autocomplete?q=term` JSON API for fast autocomplete search against foreign key tables - it searches against their label column or their primary key and switches to just a prefix search against the first primary key (for speed) if the label column check takes more than 500ms. A new `/-/debug/autocomplete` page lets you try this out.
- A `<datasette-autocomplete>` Web Component that uses that API.
- Table pages now get an insert button above the table, and little edit and delete icons next to each row. All three trigger custom modal dialogs. The edit/insert dialog is a full form - the delete one is just confirmation.
- A new `/<database>/<table>/-/fragment?_row=` endpoint which returns a rendered fragment of HTML for the specified row. This is used by the insert/edit code to partially update the table to reflect those changes. Uses a new `data-row="{{ row.row_path }}"` attribute on the `<tr>` to enable the replacement.
- A new default column type called `textarea` which users can use to specify a multi-line textarea for a column
- A new JavaScript plugin hook, [makeColumnField()](https://github.com/simonw/datasette/blob/3f7d389caf72081cb7d6d5c6b25e9a19068e744e/docs/javascript_plugins.rst#makecolumnfieldcontext), which plugins can use to add custom form fields to the edit form. Datasette [uses this itself](https://github.com/simonw/datasette/blob/3f7d389caf72081cb7d6d5c6b25e9a19068e744e/datasette/static/table.js#L1181-L1209) for the JSON field to add client-side JSON validation. I iterated a *lot* on this one, including spinning up a `datasette-prosemirror` plugin and a branch of `datasette-files` to fully exercise it.

----
📚 Documentation preview 📚: https://datasette--2781.org.readthedocs.build/en/2781/

https://github.com/user-attachments/assets/2c18b8a4-975f-4c7b-9573-ec6040fe8223


simonw:
There's one significant missing feature: you can't use any of this stuff on the row page yet.

I could add things to the row actions menu but that would be visually inconsistent with how the table page works, so I'm tempted to go with the icons instead.

simonw:
... actually row pages don't display the primary key column like the table page, and the fact that there's a table on the row page with only one row in it feels a bit weird. I'm going to do the action menu.

simonw:
Let's do it in the actions menu, but those actions trigger the modal dialogs - but when the modal dialogs finish we refresh the page (for edit) or redirect back to the table itself with a flash message (for delete)

simonw:
I'm going to refactor the dialogs out of `table.js` and into `edit-tools.js`

simonw:
Yeah the row action buttons feel pretty good:

row-actions

PR #2785, opened 2026-06-16
Test Datasette JavaScript features using Playwright

### LINKED ISSUE #2779: Spin up a Playwright test suite
It's time to have proper Playwright browser automation tests - Datasette's JavaScript is getting complex enough that it's worth it.

I'll have these run in CI as a separate task, against just the most recent Python version but with a matrix of Playwright test browsers.

simonw:
Refs:
- #2779 
- #2781 

Includes setup to run Playwright tests in CI using Python 3.14 only (no benefit to testing in multiple Pythons) and with a matrix to run Chromium, WebKit and Firefox.

----
📚 Documentation preview 📚: https://datasette--2785.org.readthedocs.build/en/2785/

Mostly written by Codex Desktop and GPT-5.5 xhigh.


tests/test_playwright.py
```diff
@@ -0,0 +1,669 @@
...
+    page.evaluate("""
+    () => {
+      window._datasetteTableData = {
+        database: "data",
+        table: "projects",
+        tableUrl: "/data/projects",
+      };
+    }
+    """)
+    page.add_script_tag(url=f"{datasette_server}-/static/edit-tools.js")
```
simonw: I'm going to use a real fixture table page rather than hack the script in like this.

simonw:
> Looking at `tests/test_playwright.py` it's clear these tests were ported from Node tests that had no DOM - consider how these tests could be updated to actually exercise the features, not just directly run things through JavaScript on the page

Issue #2788, opened 2026-06-17
UI (and API) for altering a table


simonw:
A natural pairing for:
- #2787 

We already depend on `sqlite-utils` which has robust mechanisms for this.

This feature will make [datasette-edit-schema](https://github.com/simonw/datasette-edit-schema) obsolete.

Issue #2787, opened 2026-06-17
UI for creating a table


simonw:
Now that we have insert/edit/delete in the UI it feels like creating a table is a missing feature, follows:
- #2781

simonw:
Marathon Codex Desktop session: https://gist.github.com/simonw/dc8b19f7df4846e58aefc770377e9e35 - now in PR:

- #2789

PR #2789, opened 2026-06-17
Create table, alter table - APIs and modals

simonw:
Refs:
- #2788 
- #2787 

Video demo (before a few final cosmetic fixes):

https://github.com/user-attachments/assets/4976b33e-b218-4177-a969-c25f316a1052

----
📚 Documentation preview 📚: https://datasette--2789.org.readthedocs.build/en/2789/


datasette/static/app.css
```diff
@@ -1749,6 +1749,788 @@ datasette-autocomplete input[type="text"],
...
+    animation: datasette-modal-slide-in var(--modal-animation-duration, 0.2s) ease-out;
+    overflow: hidden;
+    font-family: system-ui, -apple-system, sans-serif;
+    background: var(--card);
+}
+
+dialog.table-create-dialog[open] {
+    display: flex;
+    flex-direction: column;
+}
```
simonw, line 1776: This is consistent with our other dialogs but there's a whole lot of duplicate model code now. Filing an issue to clean that up later:

- https://github.com/simonw/datasette/issues/2790

simonw:
Pyodide tests are failing with:

> `ValueError: Couldn't find a pure Python 3 wheel for 'pydantic-core==2.46.4'. You can use `micropip.install(..., keep_going=True)` to get a list of all packages with missing wheels.`

That shouldn't be an issue now, see: https://simonwillison.net/2026/Jun/13/publishing-wasm-wheels/

Might need to upgrade Pyodide (here and in Datasette Lite).

simonw:
There's a feature missing from both create and alter: foreign key constraints. These are worth supporting, particularly since the edit/insert modals have neat support for them.

Related to that: setting the "label column" for a table. Might need a larger issue for that since we need to store label columns somewhere, probably in the same place as custom column types.

datasette/views/table_create_alter.py
```diff
@@ -25,6 +33,7 @@
     sqlite_type: column_type
     for column_type, sqlite_type in CREATE_TABLE_SQLITE_TYPES.items()
 }
+TABLE_NAME_RE = re.compile(r"^(?!sqlite_)[^\n]+$")
```
simonw, line 43: Because SQLite preserves that prefix for internal use:

```bash
python3 << 'EOF'
import sqlite3

conn = sqlite3.connect(":memory:")
cur = conn.cursor()

# Attempt 1: plain CREATE TABLE with the reserved-looking name
try:
    cur.execute("CREATE TABLE sqlite_misx (id INTEGER PRIMARY KEY, name TEXT)")
    print("Attempt 1 (plain CREATE TABLE): SUCCESS")
except sqlite3.Error as e:
    print(f"Attempt 1 (plain CREATE TABLE): FAILED -> {type(e).__name__}: {e}")

conn.close()
EOF
```

datasette/static/edit-tools.js
```diff
@@ -43,102 +47,2820 @@ function hideRowMutationStatus() {
...
+    return "floating point number";
+  }
+  if (type === "real") {
+    return "floating point number";
+  }
+  if (type === "blob") {
+    return "blob - binary data";
+  }
+  return type;
+}
```
simonw, line 76: I'm going to make this an object instead.

datasette/views/row.py
```diff
@@ -212,6 +212,7 @@ async def template_data():
                     table,
                     not is_table,
                     None,
+                    None,
                 ),
```
simonw, line 216: Hard to read

simonw:
Alter table doesn't currently let you set foreign keys. Create table doesn't let you set a default expression.

I had Codex review them for other differences:

> I compared the running forms and checked the shared JS.
> 
> **Create Table**
> - Creates a brand new table; has a **Table name** field.
> - Starts with `id` integer primary key plus one blank text column.
> - Per column: name, SQLite type, move/remove, advanced options.
> - Advanced options: custom Datasette column type, foreign key target, primary key.
> - Foreign keys are exposed in the UI, but only for non-first columns.
> - Primary key UI is effectively first-column-only.
> - Submits immediately with **Create table**; no review step.
> - No UI for `not null`, default value, or default expression.
> 
> **Alter Table**
> - Edits an existing table; no table name field.
> - Preloads every existing column.
> - Per column: rename, change SQLite type, reorder, drop/remove, advanced options.
> - Advanced options: custom Datasette column type, `Not null`, default value, default expression, primary key.
> - Can add new columns and change/drop/reorder existing ones.
> - Has **Review changes** then **Apply changes**, with a warning for dropped columns.
> - Includes **Drop table** in the same modal.
> - No foreign-key editing control in the form right now.
> 
> So the big asymmetry is: create has foreign-key selection, alter has defaults/not-null/review/drop-table. They share the column row mechanics, type/custom-type controls, move behavior, and primary-key ordering rules.

Need to unify the code and features before landing this.

simonw:
Let's hide the default value stuff in a summary/details thing on both UIs - change text to "or default to a specific value" - and in the expression menu extend it to say things like "Current timestamp in uTC, e.g. 2026-05-01 13:34:00".

datasette/views/table_create_alter.py
```diff
@@ -0,0 +1,1202 @@
...
+    return options
+
+
+SqliteApiType = Literal["text", "integer", "float", "blob"]
+DefaultExpr = Literal["current_timestamp", "current_date", "current_time"]
+DEFAULT_EXPR_SQL = {
+    "current_timestamp": "CURRENT_TIMESTAMP",
+    "current_date": "CURRENT_DATE",
+    "current_time": "CURRENT_TIME",
+}
```
simonw, line 341: Let's add unix timestamp options here as well.

simonw: - `current_unixtime` - integer seconds since the epoch
- `current_unixtime_ms` - integer ms since the epoch

simonw:
Found this `sqlite-utils` bug while trying this out:

- https://github.com/simonw/sqlite-utils/issues/759

CleanShot 2026-06-22 at 11 49 38@2x

I'm going to fix that in `sqlite-utils` rather than working around it here.

simonw:
If you sort by a column, then rename that column using alter table, you get redirected back to a page with a 400 error saying `Cannot sort table by old_name_of_column`.

Instead it should remove that `_sort=` before the redirect.

simonw:
The alter table form needs to let you rename a table.

simonw:
In create table in the foreign key list we should show ALL foreign keys, not just the ones that match the selected SQLite column type - then if the user picks a foreign key that is of a different column type we should switch the column type in that select box to match the selected foreign key.

simonw:
When selecting a foreign key column, if the column name has not yet been set, set the column name to `table_id` e.g. if the foreign key is to `pypi_versions.id` then the column name should be set to `pypi_versions_id`.

simonw:
Here's a video of everything so far:


https://github.com/user-attachments/assets/4976b33e-b218-4177-a969-c25f316a1052

simonw:
The video shows that in macOS Safari the select boxes are not the same height as the input boxes.

simonw:
That fixed the Safari select problem:

CleanShot 2026-06-22 at 12 32 10@2x

Issue #2002, opened 2023-01-24
Document how actors are displayed


simonw:
https://github.com/simonw/datasette/blob/e4ebef082de90db4e1b8527abc0d582b7ae0bc9d/datasette/utils/__init__.py#L1052-L1056

This logic should be reflected in the documentation on https://docs.datasette.io/en/stable/authentication.html#actors

PR #2792, opened 2026-06-18
Document how actors are displayed

cycsmail:
Closes:
- #2002

Datasette uses `display_actor()` to choose a short label for the current actor in places like the navigation menu and the `/-/logout` page, but that behaviour wasn't documented anywhere. This adds a "How actors are displayed" section to the authentication docs that spells out the `display` / `name` / `username` / `login` / `id` key order it looks through and the `str(actor)` fallback when none are set.

I also added a test in `tests/test_docs.py` that reads the key tuple straight from `display_actor` and asserts each key is mentioned in the new docs section, so the docs and the code can't quietly drift apart. The test_docs.py tests pass, and the new test fails if the docs section is removed.

### DIFF
--- a/tests/test_docs.py
+++ b/tests/test_docs.py
@@ -171,6 +171,24 @@ def test_functions_marked_with_documented_are_documented(documented_labels, subt
             assert fn._datasette_docs_label in documented_labels
 
 
+def test_display_actor_logic_is_documented():
+    # The keys used by datasette.utils.display_actor should be reflected
+    # in the authentication documentation - see issue #2002
+    content = (docs_path / "authentication.rst").read_text()
+    section = content.split(".. _authentication_actor_display:")[-1].split(
+        ".. _authentication_root:"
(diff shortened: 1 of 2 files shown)


tests/test_docs.py
```diff
@@ -171,6 +171,24 @@ def test_functions_marked_with_documented_are_documented(documented_labels, subt
...
+    # The keys used by datasette.utils.display_actor should be reflected
+    # in the authentication documentation - see issue #2002
+    content = (docs_path / "authentication.rst").read_text()
+    section = content.split(".. _authentication_actor_display:")[-1].split(
+        ".. _authentication_root:"
+    )[0]
+    for key in utils.display_actor.__code__.co_consts:
+        if isinstance(key, tuple):
+            display_keys = key
+            break
```
simonw: Bit of a weird test, I don't think we need this.

PR #2794, opened 2026-06-20
Add create table execute-write template

simonw:
The `/db/-/execute-write` page now has a "Create table" template I addition to the templates for insert row / update rows / delete rows. It looks like this:

create-table

Also fixed it so if you select a template it no longer refreshes the entire page and closes the template expandable on you.

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -625,9 +625,12 @@ class Datasette:
                 return action
         return None
 
-    async def refresh_schemas(self):
+    async def refresh_schemas(self, *, force=False):
         # Throttle schema refreshes to at most once per second
-        if time.monotonic() - getattr(self, "_last_schema_refresh", 0) < 1.0:
+        if (
+            not force
+            and time.monotonic() - getattr(self, "_last_schema_refresh", 0) < 1.0
+        ):
             return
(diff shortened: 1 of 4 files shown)


datasette/app.py
```diff
@@ -625,9 +625,12 @@ def get_action(self, name_or_abbr: str):
                 return action
         return None
 
-    async def refresh_schemas(self):
+    async def refresh_schemas(self, *, force=False):
```
simonw, line 628: We had a bug where you could add a new table but it would not be represented in the list of tables (for insert/etc) until you refreshed the page again. This was a timing issue with the schema refreshing - the new `force=True` parameter lets us fix that.

tests/test_queries.py
```diff
@@ -1851,6 +1883,136 @@ async def test_execute_write_templates_are_filtered_by_permission_and_server_gen
...
+        ")\n"
+        "values (\n"
+        "  :name\n"
+        ")"
+    )
+    assert await db.table_exists("selectable")
+
+
+@pytest.mark.asyncio
+async def test_execute_write_create_table_template_optional_created_column_is_valid():
```
simonw: Pointless test.

PR #2810, opened 2026-06-26
Show facet option for the first column on view pages

### LINKED ISSUE #2045: First column on a view page has no facet option in cog menu
e.g. first column on this page - cog menu has no option to facet.

https://datasette.io/content/tools

image

cycsmail:
Closes #2045

The cog menu was hiding "Facet by this" on the first column of a view. The check used a positional `th:first-of-type` heuristic to skip the special row-link column, but on a view the first column is a real data column so it got skipped too.

Switched it to key off the `data-is-link-column` attribute the template already sets for the special link column (the same thing `hasMultipleVisibleColumns` and the mobile column menu already use), so the link column is still skipped but a view's first column now gets the facet option.

Added a test that drives `buildColumnActionItems` via node against a minimal fake DOM: the first column of a view now offers faceting, while the link column and a single pk still don't.

### DIFF
--- a/datasette/static/table.js
+++ b/datasette/static/table.js
@@ -379,7 +379,7 @@ function buildColumnActionItems(manager, th, options) {
   var column = th.dataset.column;
   var columnActions = [];
   var isSortable = !!th.querySelector("a");
-  var isFirstColumn = th.parentElement.querySelector("th:first-of-type") === th;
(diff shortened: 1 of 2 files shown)


cycsmail:
just checking in on this one, all checks are green. happy to rebase or tweak anything if that makes it easier to review.

PR #2812, opened 2026-06-29
Handle recursive CTEs in query analysis, refs #2809

### LINKED ISSUE #2809: Save query feature doesn't work with SQLite recursive queries
A few bugs here:

1. You can't save a recursive query because this happens:
    Image
2. The submit button is disabled but there's no visual indication of that, it just stops workning:
    Image
3. If you re-enable that button, submitting it fails without an error message

simonw:
SQLite emits a SQLITE_RECURSIVE authorizer callback while compiling recursive CTEs. Datasette was treating that bookkeeping callback as an unknown operation, which made saved-query analysis disable otherwise valid read-only queries.

Ignore the callback while preserving the real table reads, and cover the saved-query create/analyze flow with a recursive CTE regression test.

----
📚 Documentation preview 📚: https://datasette--2812.org.readthedocs.build/en/2812/

### DIFF
--- a/datasette/utils/sql_analysis.py
+++ b/datasette/utils/sql_analysis.py
@@ -150,8 +150,6 @@ _SQLITE_INTERNAL_SCHEMA_FUNCTIONS = {
     "sqlite_rename_test",
     "substr",
 }
-_SQLITE_RECURSIVE = getattr(sqlite3, "SQLITE_RECURSIVE", None)
-
 _AUTHORIZER_ACTION_NAMES = {
     getattr(sqlite3, name): name
     for name in (
@@ -392,7 +390,7 @@ def analyze_sql_tables(
             )
             return sqlite3.SQLITE_OK
 
-        if _SQLITE_RECURSIVE is not None and action == _SQLITE_RECURSIVE:
+        if action == sqlite3.SQLITE_RECURSIVE:
             # Recursive CTE bookkeeping; table reads are reported separately.
             return sqlite3.SQLITE_OK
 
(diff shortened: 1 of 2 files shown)


datasette/utils/sql_analysis.py
```diff
@@ -391,6 +392,10 @@ def authorizer(action, arg1, arg2, sqlite_schema, source):
             )
             return sqlite3.SQLITE_OK
 
+        if _SQLITE_RECURSIVE is not None and action == _SQLITE_RECURSIVE:
```
simonw: Why would `_SQLITE_RECURSIVE` be None?

simonw: That constant is present in Python 3.10 through 3.14:
```bash
uv run --python 3.10 python -c "import sqlite3; print(sqlite3.SQLITE_RECURSIVE)"
```
```
Python 3.10.20: 33
Python 3.11.15: 33
Python 3.12.13: 33
Python 3.13.12: 33
Python 3.14.3: 33
```

simonw: Code in cPython 10 years ago:

https://github.com/python/cpython/blame/b67cbbda3a022cec5e2ad929f0531162166e7c8d/Modules/_sqlite/module.c#L314

```
#if SQLITE_VERSION_NUMBER >= 3008003
    {"SQLITE_RECURSIVE", SQLITE_RECURSIVE},
```
But https://github.com/python/cpython/commit/00b1ae0dfcd3ef7de428d8d8d8aca819938ababa lists SQLite 3.8.3 for that constant.

PR #2813, opened 2026-06-29
Bulk insert mechanism in the insert row dialog

simonw:
- Click "Insert multiple rows" from the insert dialog
- Paste in JSON or TSV or CSV - or drop on a file
- Preview table showing what will be inserted
- Rows get inserted via the JSON API
- If data includes primary keys and user has `update-rows` permission they can upsert instead
- Same UI is also available on create table, as "create table from rows", using existing API.

It's effectively https://github.com/datasette/datasette-import as a core feature.

### DIFF
--- a/datasette/views/table_create_alter.py
+++ b/datasette/views/table_create_alter.py
@@ -267,6 +267,11 @@ async def _create_table_ui_context(
         "databaseName": database_name,
         "columnTypes": CREATE_TABLE_COLUMN_TYPES,
         "defaultExpressions": default_expression_options(),
+        "canInsertRows": await datasette.allowed(
+            action="insert-row",
+            resource=DatabaseResource(database=database_name),
+            actor=request.actor,
+        ),
     }
     can_set_column_type = await datasette.allowed(
         action="set-column-type",
--- a/datasette/static/app.css
+++ b/datasette/static/app.css
@@ -1819,20 +1819,30 @@ textarea.row-edit-input {
     display: none;
 }
 
-.row-edit-bulk-label {
-    color: var(--ink);
-    font-family: ui-monospace, SFMono-Regular, Menlo, Consolas, monospace;
-    font-size: 0.82rem;
-}
-
 .row-edit-bulk-actions {
     display: flex;
-    justify-content: flex-end;
+    align-items: center;
+    flex-wrap: wrap;
+    gap: 8px;
+    justify-content: flex-start;
 }
 
 .row-edit-bulk-actions .btn {
     padding-left: 12px;
     padding-right: 12px;
+}
+
+.row-edit-copy-template-label-narrow {
+    display: none;
+}
+
+.row-edit-bulk-template-note {
+    color: var(--muted);
+    font-size: 0.82rem;
+}
+
+.row-edit-bulk-template-note-narrow {
+    display: none;
 }
 
 .row-edit-bulk-textarea {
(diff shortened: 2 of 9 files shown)


simonw:
Insert screen now looks like this:

CleanShot 2026-06-29 at 15 13 43@2x

simonw:
Point of confusion: I tried importing the same external data (with a concrete primary key - it was the JSON from https://api.github.com/users/simonw/repos) twice - once to create the table, and then again with the edit feature.

This happened:

CleanShot 2026-06-29 at 17 46 39@2x

I guess I was actually trying to do an upsert here, which the UI doesn't yet support.

Need to think about if upsert should be a feature and how it should work.

simonw:
Note that upsert requires an additional permission, `update-row` - not just `insert-row`.

simonw:
Bug reports highlighted by GPT-5.5 xhigh when it was recording [this video](https://simonwillison.net/2026/Jun/30/shot-scraper-video/):

<details><summary>Bugs found in `bulk-insert`</summary>

## 1. Bulk insert overrides SQLite defaults for omitted columns

Bulk insert currently materializes every insertable column in the row payload, even when the pasted data omitted that column.

The relevant code is in `datasette/static/edit-tools.js`:

- `rowObjectForBulkInsert()` iterates through every column.
- `normalizeBulkInsertCell()` turns missing values into `null`, or `""` for `NOT NULL` columns.

That means a pasted subset like:

```csv
title
Only title
```

against a table like:

```sql
create table tasks (
  id integer primary key,
  title text not null,
  status text not null default 'todo',
  score integer default 5
);
```

gets sent to the API roughly as:

```json
{"title": "Only title", "status": "", "score": null}
```

This prevents SQLite defaults from firing. If the payload omitted `status` and `score`, SQLite would store `status = 'todo'` and `score = 5`.

Suggested fix: preserve the distinction between an explicitly blank pasted cell and a column that was absent from the pasted input. For absent columns, omit the key from the row payload so SQLite defaults can apply.

## 2. Create-table-from-CSV loses numeric inference when numeric columns contain blanks

The create-table-from-data flow infers numeric column types while ignoring blank cells, but then serializes blank numeric cells as empty strings.

The relevant code is in `datasette/static/edit-tools.js`:

- `inferCreateTableDelimitedColumnType()` ignores blank values when deciding `integer` or `float`.
- `coerceCreateTableDelimitedValue()` returns `""` for blank cells, even if the inferred type is numeric.

For example:

```csv
name,score
A,1
B,
```

is inferred as having a numeric `score` column, but the preview rows sent to the create-table API contain:

```json
[
  {"name": "A", "score": 1},
  {"name": "B", "score": ""}
]
```

`sqlite-utils` sees the empty string and creates `score` as `TEXT`, not `INTEGER`.

Suggested fix: when a column has been inferred as `integer` or `float`, convert blank cells to `null` instead of `""`. Alternatively, send explicit column type information to the API rather than relying on `sqlite-utils` to infer types from mixed values.

</details>

simonw:
Codex also spotted:

> P2: “Create table from data” is exposed with only create-table permission. `_create_table_ui_context` (line 260) gates the dialog on create-table, and the link is always rendered at `edit-tools.js` (line 2193), but submitting rows hits the server check requiring insert-row at `table_create_alter.py` (line 833). Users with create-only permission get a UI path that fails after preview.

simonw:
Last feature before landing this: upset support. We'll detect if the user has `update-rows` permission and the data they are bulk-inserting includes values for the primary keys and if so we'll let them pick between ignore existing rows, quit on existing rows or upsert existing rows.

simonw:
There's a video demo generated using `shot-scraper video` in this blog post: https://simonwillison.net/2026/Jun/30/shot-scraper-video/

datasette/views/table_create_alter.py
```diff
@@ -267,6 +267,11 @@ async def _create_table_ui_context(
         "databaseName": database_name,
         "columnTypes": CREATE_TABLE_COLUMN_TYPES,
         "defaultExpressions": default_expression_options(),
+        "canInsertRows": await datasette.allowed(
+            action="insert-row",
+            resource=DatabaseResource(database=database_name),
+            actor=request.actor,
```
simonw, line 273: This is `DatabaseResource` and not `TableResource` because here we are in the table creation UI for the "create table from rows" - we can't check permission on the table because we have not created it yet.

Issue #1950, opened 2022-12-13
Bad ?_sort returns a 500 error, should be a 400


simonw:
https://latest.datasette.io/fixtures/facetable?_sort=bad

<img width="453" alt="image" src="https://user-images.githubusercontent.com/9599/207454481-2ba2e2bd-7400-47a6-b8a1-f898a2d364d8.png">

simonw:
https://github.com/simonw/datasette/blob/d4b98d3924dec625a99236e65b1b169ff957381f/datasette/views/table.py#L392-L400

simonw:
Most places I use that exception at the moment set their own non-500 status error:
```
datasette % rg DatasetteError
datasette/handle_exception.py
7:from .views.base import DatasetteError
33:        elif isinstance(exception, DatasetteError):

datasette/filters.py
2:from datasette.views.base import DatasetteError
22:                raise DatasetteError("_where= is not allowed", status=403)
141:                    raise DatasetteError(

datasette/views/table.py
34:from .base import BaseView, DataView, DatasetteError, ureg, _error
178:                raise DatasetteError(
192:                raise DatasetteError(
390:            raise DatasetteError("Cannot use _sort and _sort_desc at the same time")
394:                raise DatasetteError(f"Cannot sort table by {sort}")
400:                raise DatasetteError(f"Cannot sort table by {sort_desc}")

datasette/views/base.py
39:class DatasetteError(Exception):
219:            raise DatasetteError(str(e), title="Invalid SQL", status=400)
222:            raise DatasetteError(str(e))
224:        except DatasetteError:
382:            raise DatasetteError(
402:            raise DatasetteError(str(e), title="Invalid SQL", status=400)
405:            raise DatasetteError(str(e))
407:        except DatasetteError:

datasette/views/table2.py
28:from .base import DataView, DatasetteError, ureg
296:                raise DatasetteError(
310:                raise DatasetteError(
472:            raise DatasetteError("Cannot use _sort and _sort_desc at the same time")
476:                raise DatasetteError(f"Cannot sort table by {sort}")
482:                raise DatasetteError(f"Cannot sort table by {sort_desc}")

datasette/views/database.py
31:from .base import BaseView, DatasetteError, DataView, _error
188:            raise DatasetteError("Invalid database", status=404)
190:            raise DatasetteError("Cannot download in-memory databases", status=404)
194:            raise DatasetteError("Cannot download database", status=404)
```

PR #2815, opened 2026-06-30
Return 400 instead of 500 for wrong-arity composite-PK row URLs (#2811)

### LINKED ISSUE #2811: Composite-primary-key row pages return HTTP 500 (unbound bind parameter) on a malformed PK URL, should be 400
A `GET` for a **row page** on a table with a **composite (multi-column) primary key** returns **HTTP 500** (`sqlite3.ProgrammingError`, unbound bind parameter) when the URL has the wrong number of comma-separated primary-key components — most commonly when the separating comma arrives percent-encoded as `%2C` (proxies and crawlers routinely re-encode reserved characters in paths).

Single-PK tables don't hit this: a non-canonical single-PK URL 404s, and `handle_404`'s `%`→tilde redirect recovers it to the canonical form. The identical pk-count mismatch is already guarded with `BadRequest` in `_fragment_request_for_row`, but the row-page resolution path has no such guard — so it raises an uncaught `ProgrammingError` (500) before the 404 handler can run.

### Reproduce (confirmed on 0.65.2 and 1.0a35)

```
sqlite3 demo.db "create table t (a text, b text, primary key (a, b)); insert into t values ('x', 'y');"
datasette demo.db
curl -s -o /dev/null -w '%{http_code}\n' 'http://127.0.0.1:8001/demo/t/x,y'     # 200
curl -s -o /dev/null -w '%{http_code}\n' 'http://127.0.0.1:8001/demo/t/x%2Cy'   # 500  <-- bug
```

Server log for the `%2C` request:

```
ERROR: ... sql = 'select * from t where "a"=:p0 AND "b"=:p1', params = {'p0': 'x%2Cy'}: You did not supply a value for binding parameter :p1.
sqlite3.ProgrammingError: You did not supply a value for binding parameter :p1.
```

### Root cause

The row resolver builds one `:pN` placeholder per primary-key column but binds only the supplied components. When `%2C` collapses the comma-separated arity (one value where the table has two PK columns), the trailing `:p1` is left unbound → `ProgrammingError` (a 500) at execute time, *before* any 404 — so the existing `handle_404` recovery never runs.

### Suggested fix

Guard the pk-value count against the pk-column count in the row resolver, raising `BadRequest`, mirroring the existing `_fragment_request_for_row` check:

```python
if len(pk_values) != len(pks):
    raise BadRequest("Invalid primary key value count for this table")
```

A malformed-arity row URL is a client error, so **400** is the appropriate status — consistent with `_fragment_request_for_row` and with the `?_sort` errors → 400 change (#1950).

HarperZ9:
Fixes #2811

## The bug

A `GET` for a row page on a table with a composite (multi-column) primary key returns **HTTP 500** (`sqlite3.ProgrammingError`, unbound bind parameter) when the URL has the wrong number of comma-separated primary-key components. The most common trigger in the wild is a proxy/crawler re-encoding the separating comma as `%2C`, which collapses the arity.

### Reproduce (confirmed on 1.0a35 / main `34ab85e`)

```console
sqlite3 demo.db "create table t (a text, b text, primary key (a, b)); insert into t values ('x', 'y');"
datasette demo.db
curl -s -o /dev/null -w '%{http_code}\n' 'http://127.0.0.1:8001/demo/t/x,y'     # 200
curl -s -o /dev/null -w '%{http_code}\n' 'http://127.0.0.1:8001/demo/t/x%2Cy'   # 500  <-- bug
curl -s -o /dev/null -w '%{http_code}\n' 'http://127.0.0.1:8001/demo/t/x'       # 500  <-- bug
```

## Root cause

`row_sql_params_pks` (`datasette/utils/__init__.py`) builds one `:pN` bind placeholder **per primary-key column** but binds params only for the **supplied** URL components:

```python
wheres = [f'"{pk}"=:p{i}' for i, pk in enumerate(pks)]   # one per PK column
...
for i, pk_value in enumerate(pk_values):                  # one per supplied value
    params[f"p{i}"] = pk_value
```

When too few components are supplied, the trailing placeholder (e.g. `:p1`) is left unbound and `Datasette.resolve_row` (`datasette/app.py`) executes the SQL with no arity guard, so SQLite raises `You did not supply a value for binding parameter :p1` - a 500, raised before the existing `handle_404` recovery can run. When too many components are supplied, the extra params are silently ignored and a misleading result is returned.

The identical pk-count mismatch is **already guarded** with `BadRequest` in `_fragment_request_for_row` (`datasette/views/table.py`), but the row-page resolution path had no such guard.

## The fix

Add a primary-key arity guard in `resolve_row`, mirroring the existing convention in `datasette/views/table.py`:

```python
sql, params, pks = await row_sql_params_pks(db, table_name, pk_values)
if len(pk_values) != len(pks):
    raise BadRequest(
        "URL row identifier does not match the primary key for this table"
    )
```

A malformed-arity row URL is a client error, so **400** is the appropriate status, consistent with `_fragment_request_for_row` and the `?_sort` errors -> 400 change (#1950).

`BadRequest` was not previously imported into `app.py`, so it is added to the `from .utils.asgi import (...)` block. It is a `Base400` subclass that yields HTTP 400.

After the fix:

```console
/demo/t/x,y    -> 200   (canonical, unchanged)
/demo/t/x%2Cy  -> 400   (was 500)
/demo/t/x      -> 400   (was 500)
/demo/t/x,y,z  -> 400   (was a silent wrong-result 200)
```

A single-column PK value that legitimately contains a comma is unaffected: it tilde-encodes (`a,b` -> `a~2Cb`), so it stays one component and resolves normally.

## Tests

Added `test_row_pk_arity_mismatch_returns_400` (parametrized over too-few / too-many components and HTML / `.json`) plus `test_row_compound_pk_correct_arity` in `tests/test_api.py`, using the existing `compound_primary_key` fixture. The new tests fail on `main` (500 / silent-200) and pass with the fix. The 400 assertion style follows the existing `test_sort_errors` precedent in `tests/test_table_html.py`.

Current local verification on 2026-07-02 with Python 3.12.10:

```console
uv run --group dev python -m pytest tests/test_api.py::test_row_pk_arity_mismatch_returns_400 tests/test_api.py::test_row_compound_pk_correct_arity -q
5 passed

uv run ruff check datasette/app.py tests/test_api.py
All checks passed!

git diff --check
# clean
```

---

Authored with AI assistance (Claude); reviewed and verified before submitting (bug reproduced before the change, fix and tests verified after).

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -132,6 +132,7 @@ from .utils import (
 )
 from .utils.asgi import (
     AsgiLifespan,
+    BadRequest,
     Forbidden,
     NotFound,
     DatabaseNotFound,
@@ -2789,6 +2790,10 @@ class Datasette:
         db, table_name, _ = await self.resolve_table(request)
         pk_values = urlsafe_components(request.url_vars["pks"])
         sql, params, pks = await row_sql_params_pks(db, table_name, pk_values)
+        if len(pk_values) != len(pks):
+            raise BadRequest(
+                "URL row identifier does not match the primary key for this table"
+            )
         results = await db.execute(sql, params, truncate=True)
         row = results.first()
         if row is None:
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -372,6 +372,40 @@ async def test_row(ds_client):
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize("suffix", ("", ".json"))
+@pytest.mark.parametrize(
+    "row_path",
+    (
+        "a",  # too few components for a two-column primary key
+        "a,b,c",  # too many components for a two-column primary key
+    ),
+)
+async def test_row_pk_arity_mismatch_returns_400(ds_client, row_path, suffix):
+    # A row URL with the wrong number of comma-separated primary key
+    # components used to raise an uncaught sqlite3.ProgrammingError (HTTP 500)
+    # because the SQL had one bind placeholder per PK column but params were
+    # only bound for the supplied components. It should be a 400 instead,
+    # mirroring the existing guard in datasette/views/table.py.
+    response = await ds_client.get(
+        "/fixtures/compound_primary_key/{}{}".format(row_path, suffix)
+    )
+    assert response.status_code == 400
+    if suffix == ".json":
+        assert response.json()["ok"] is False
+        assert response.json()["status"] == 400
+
+
+@pytest.mark.asyncio
+async def test_row_compound_pk_correct_arity(ds_client):
+    # The valid two-component URL still resolves the row.
+    response = await ds_client.get(
+        "/fixtures/compound_primary_key/a,b.json?_shape=objects"
+    )
+    assert response.status_code == 200
+    assert response.json()["rows"] == [{"pk1": "a", "pk2": "b", "content": "c"}]
+
+
+@pytest.mark.asyncio
 async def test_row_strange_table_name(ds_client):
     response = await ds_client.get(
         "/fixtures/table~2Fwith~2Fslashes~2Ecsv/3.json?_shape=objects"


HarperZ9:
Fresh local verification for this PR, from a Windows checkout of head `aee8f0ca5b3de5770886d72c4d704c9f800bc823` with Python 3.12.10:

- `uv run --group dev python -m pytest tests/test_api.py::test_row_pk_arity_mismatch_returns_400 tests/test_api.py::test_row_compound_pk_correct_arity -q` -> 5 passed
- `uv run ruff check datasette/app.py tests/test_api.py` -> all checks passed
- `git diff --check` -> clean

This keeps verification focused on the composite-primary-key row URL arity guard and the touched files.

PR #2818, opened 2026-07-02
Support CREATE VIEW / DROP VIEW in execute-write-sql

simonw:
Refs:
- #2819 

<details><summary>Opus 4.8 PR</summary>

CREATE VIEW and DROP VIEW were being rejected as "Unsupported SQL
operation" because decision_for_write_sql_operation() only had cases
for target_type "table" and "index". Add dedicated create-view and
drop-view permission actions and wire them into the write-SQL
decision logic.

Also fix a related bug in analyze_sql_tables(): SQLite's authorizer
fires a spurious SQLITE_DELETE callback against the view name when a
view is dropped (the same thing it does for dropped tables), which
was incorrectly surfaced as a delete-row requirement on the view.
Broaden the existing drop-table-delete suppression to cover dropped
views too.

</details>

### DIFF
--- a/datasette/default_actions.py
+++ b/datasette/default_actions.py
@@ -62,6 +62,12 @@ def register_actions():
             resource_class=DatabaseResource,
         ),
         Action(
+            name="create-view",
+            abbr="cv",
+            description="Create views",
+            resource_class=DatabaseResource,
+        ),
+        Action(
             name="store-query",
             abbr="sq",
             description="Create stored queries",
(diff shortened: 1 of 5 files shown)


datasette/default_actions.py
```diff
@@ -111,6 +117,12 @@ def register_actions():
             description="Drop tables",
             resource_class=TableResource,
         ),
+        Action(
+            name="drop-view",
+            abbr="dv",
+            description="Drop views",
+            resource_class=TableResource,
```
simonw, line 124: Note that `TableResource` in the permissions system actually covers both tables and views. I think that's OK.

PR #2824, opened 2026-07-04
Fable review of JSON API consistency and subsequent improvements

simonw:
Started this process with: https://claude.ai/code/session_01GrHZSypDfMnym1tM5XJAFZ

> Ignore existing documentation. Create a fresh detailed document describing the existing JSON API - all endpoints. Save as existing-api.md
>
> Then review that API for consistency and completeness with respect to the upcoming Datasette 1.0 stable release, save as stable-api-recommendations.md

- Existing API docs: https://github.com/simonw/datasette/blob/8985ecf4387164e76147e7188693a08dc1a7dd1e/existing-api.md
- Stable API recommendations: https://github.com/simonw/datasette/blob/8985ecf4387164e76147e7188693a08dc1a7dd1e/stable-api-recommendations.md

----
📚 Documentation preview 📚: https://datasette--2824.org.readthedocs.build/en/2824/


datasette/views/base.py
```diff
@@ -49,9 +50,7 @@ async def method_not_allowed(self, request):
             request.path.endswith(".json")
             or request.headers.get("content-type") == "application/json"
         ):
-            response = Response.json(
-                {"ok": False, "error": "Method not allowed"}, status=405
-            )
+            response = Response.json(error_body("Method not allowed", 405), status=405)
```
simonw: Let's find a better pattern than this, repeating the status twice is ugly.

datasette/views/stored_queries.py
```diff
@@ -199,7 +208,6 @@ async def facet_items(items):
             "queries": page.queries,
             "next": page.next,
             "next_url": next_url,
```
simonw, line 212: Now that we always include `next_url` I'm going to get rid of the `?_extra=next_url` entirely.

datasette/views/table.py
```diff
@@ -950,9 +955,7 @@ async def _validate_data(self, request, db, table_name, pks, upsert):
         def _errors(errors):
             return None, errors, {}
 
-        if not request.headers.get("content-type").startswith("application/json"):
-            # TODO: handle form-encoded data
-            return _errors(["Invalid content-type, must be application/json"])
+        # The body is parsed as JSON regardless of the Content-Type header
```
simonw, line 958: We were being inconsistent in whether we checked for `application/json` or not - the reason to check that is to avoid sneaky CSRF attacks using HTML forms and `<input type="hidden" name='{"malicious-json": "value", "ignore": "' value='"}'>` tricks, but our existing CSRF protection should prevent those from working.

So I decided NOT to check for that header on requests, since it's more convenient to just be able to stick JSON in a POST body without thinking about headers.

datasette/extras.py
```diff
@@ -113,6 +115,17 @@ def _allowed_names_for_scope(self, scope, include_internal):
             self._allowed_names[key] = names
         return names
 
+    def validate_requested(self, requested, scope):
+        """
+        Raise BadRequest if any requested extra name is not a public extra
+        for this scope. Used by data formats such as .json - HTML pages
+        silently ignore unknown names instead.
+        """
```
simonw, line 123: We run this validation on JSON but not HTML because HTML actually has a different set of available extras (since it automatically applies some), plus JSON is an API so it's worth being less lenient.

datasette/stored_queries.py
```diff
@@ -62,7 +62,6 @@ def stored_query_to_dict(query: StoredQuery) -> dict[str, Any]:
         "description_html": query.description_html,
         "hide_sql": query.hide_sql,
         "fragment": query.fragment,
-        "params": list(query.parameters),
         "parameters": list(query.parameters),
```
simonw, line 65: `params` and `parameters` were duplicates of each other.

datasette/stored_queries.py
```diff
@@ -84,7 +83,6 @@ def stored_query_page_to_dict(page: StoredQueryPage) -> dict[str, Any]:
     return {
         "queries": [stored_query_to_dict(query) for query in page.queries],
         "next": page.next,
-        "has_more": page.has_more,
```
simonw, line 87: we use `next` being `null` instead of `has_more` now.

docs/introspection.rst
```diff
@@ -7,6 +7,10 @@ Datasette includes some pages and JSON API endpoints for introspecting the curre
 
 Each of these pages can be viewed in your browser. Add ``.json`` to the URL to get back the contents as JSON.
 
+JSON responses that return an object include an ``"ok": true`` key, consistent with the rest of the :ref:`JSON API <json_api>`.
+
+The introspection endpoints documented on this page are covered by the :ref:`JSON API stability promise <json_api_stability>`, with the exception of the debug endpoints ``/-/threads``, ``/-/actions`` and ``/-/jump``, whose shapes may change in future releases.
+
```
simonw: `/-/jump` should be documented and stable, and is not a debug endpoint.

docs/json_api.rst
```diff
@@ -9,6 +9,53 @@ through the Datasette user interface can also be accessed as JSON via the API.
...
+
+      "unstable": "This API is not part of Datasette's stable interface and may change at any time"
+
+  This currently covers the instance homepage (``/.json``), the stored
+  query ``analyze``/``store``/``definition`` endpoints, ``/-/query/parameters``,
+  ``/-/execute-write/analyze`` and the JSON returned by the ``/-/permissions``
+  debug playground.
+- Debug and support endpoints are documented so you can use them, but their
+  JSON shapes are not frozen: :ref:`/-/threads <JsonDataView_threads>`,
+  :ref:`/-/actions <JsonDataView_actions>`, :ref:`/-/jump <JumpView>`,
```
simonw: Another mention of `/-/jump` as being unstable.

tests/test_permissions.py
```diff
@@ -739,6 +739,11 @@ async def test_actor_restricted_permissions(
         "path": expected_path,
     }
     expected = {
+        "ok": True,
+        "unstable": (
+            "This API is not part of Datasette's stable interface"
+            " and may change at any time"
+        ),
```
simonw: we should use a constant here

tests/test_queries.py
```diff
@@ -3742,3 +3749,105 @@ async def test_stored_write_query_with_truncated_returning_message():
...
+    assert "params" not in definition["query"]
+
+    listing = (await ds.client.get("/data/-/queries.json")).json()
+    query = [q for q in listing["queries"] if q["name"] == "with_params"][0]
+    assert query["parameters"] == ["name", "age"]
+    assert "params" not in query
+
+
+@pytest.mark.asyncio
+async def test_query_list_json_signals_pagination_via_next_only():
```
simonw: This test is pointless.

PR #2845, opened 2026-07-14
Explain permission decisions in debug tools and documentation

### LINKED ISSUE #2841: Improved documentation and tooling for permissions system
The permissions system works, and we've built a bunch of stuff against it that works, but it's still too difficult to get your head around exactly how it is implemented.

Solution is a combination of improved documentation and better interactive demo tools.

Can happen as part of:
- #2832

simonw:
Closes #2841.

## Summary

- add a structured explanation for single permission decisions, including matching rules, specificity, same-scope deny precedence, actor restrictions, required actions and default deny
- allow the `/-/check` debug tool to evaluate hypothetical actor JSON while still authorizing access as the signed-in actor
- redesign the permission debug UI around Explain, Access map, Rule explorer and Activity workflows
- document the actor/action/resource mental model, rule precedence and the purpose of each debugging tool
- add focused coverage for specificity, restrictions, action dependencies, default deny and actor input validation

## Why

The permissions implementation is powerful, but understanding an effective decision previously required correlating the Check, Rules, Allowed and Playground surfaces manually. This makes the resolver's mental model explicit and shows how each final verdict was reached.

## Validation

- `275 passed, 3 xpassed` across the focused permission suites
- full test suite: `2295 passed, 40 skipped, 6 xfailed, 15 xpassed, 141 subtests passed`; the two sandbox-blocked local server fixtures passed when rerun with local binding enabled
- Sphinx documentation build
- Black, Ruff, codespell, Cog and diff checks
- desktop and mobile browser QA for allowed and denied decisions, with no console errors

### DIFF
--- a/datasette/default_permissions/config.py
+++ b/datasette/default_permissions/config.py
@@ -96,6 +96,10 @@ class ConfigPermissionProcessor:
         """Evaluate an allow block against the current actor."""
         if allow_block is None:
             return None
+        # Values passed using ``-s permissions.* 1`` or ``0`` are parsed as
+        # integers, but should retain the CLI's boolean 1/0 behavior.
+        if isinstance(allow_block, int) and allow_block in (0, 1):
+            return bool(allow_block)
         return actor_matches_allow(self.actor, allow_block)
 
     def is_in_restriction_allowlist(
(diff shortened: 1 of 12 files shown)


simonw:
Demo of the new features:

https://github.com/user-attachments/assets/819d3301-4102-411f-9932-88771e724494

Issue #2431, opened 2024-09-23
datasette breaks on square brackets in table name


slavos1:
First off, thanks for datasette. 🙏🏻 

Checking [Evolution's](https://help.gnome.org/users/evolution/stable/) data, I've hit a snag with tables with names like `[Gmail]/Inbox`.

A minimal reproduction:

```shell
$ sqlite3 /tmp/slashes
SQLite version 3.45.1 2024-01-30 16:01:20
Enter ".help" for usage hints.
sqlite> create table "[foo]/bar" (a text);
Ctrl-D

$ datasette /tmp/slashes
Usage: datasette serve [OPTIONS] [FILES]...
Try 'datasette serve --help' for help.

Error: Connection to /tmp/slashes failed check: near "/": syntax error

$ datasette --version
datasette, version 0.64.8
```

Testing further, having only `[...]` fails with:
```shell
$ sqlite3 /tmp/slashes
SQLite version 3.45.1 2024-01-30 16:01:20
Enter ".help" for usage hints.
sqlite> create table "[foo]" (a text);

$ datasette /tmp/slashes
Usage: datasette serve [OPTIONS] [FILES]...
Try 'datasette serve --help' for help.

Error: Connection to /tmp/slashes failed check: unrecognized token: "]"
```

In summary:
| Table name | Works with datasette? |
|--------|--------|
| `[foo]/bar` | no (`failed check: near "/": syntax error`) |
| `[foo]` | no (`failed check: unrecognized token: "]"`) |
| `[foo` | yes |
| `foo]` | no (`failed check: unrecognized token: "]"`) |
| `foo/bar` | yes |

sqlite3 cli itself does not have problems with any names mentioned above.

PR #2846, opened 2026-07-14
Escape table names with [square] brackets, refs #2431

TowyTowy:
Fixes the still-reproducible part of #2431. A table whose name contains `]` (e.g. `[foo]` or `foo]`) currently crashes Datasette: several helpers quote the table name with SQLite `[bracket]` identifiers built from an f-string, and bracket quoting can't escape a `]`, giving `sqlite3.OperationalError: unrecognized token: "]"`. In practice this crashes schema introspection at startup (`internal_db.populate_schema_tables`) and 500s the table page.

These call sites now use the existing `escape_sqlite()` helper (`"double quote"` quoting with `""` escaping) — the same approach used elsewhere in the codebase and already used by `tests/test_internal_db.py` for these very PRAGMAs. Sites changed: `internal_db` `foreign_key_list`/`index_list` PRAGMAs, `get_outbound_foreign_keys`, `Database.table_counts` count query, and the default facet `select * from`.

Added a regression test (`test_table_with_reserved_characters_in_name`) parametrized over `[foo]`, `foo]`, `[foo]/bar` that asserts introspection and the table page both work.

Out of scope (happy to follow up): the `arraycontains` filter templates and the FTS-detection `LIKE` pattern.

AI-assisted (Claude), reviewed and verified locally by me — the affected test areas pass (test_api, internal_db, facets, table_api, internals_database) and Black is clean.

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -17,6 +17,7 @@ from .utils import (
     detect_fts,
     detect_primary_keys,
     detect_spatialite,
+    escape_sqlite,
     get_all_foreign_keys,
     get_outbound_foreign_keys,
     md5_not_usedforsecurity,
@@ -608,7 +609,7 @@ class Database:
             try:
                 table_count = (
                     await self.execute(
-                        f"select count(*) from (select * from [{table}] limit {self.count_limit + 1})",
+                        f"select count(*) from (select * from {escape_sqlite(table)} limit {self.count_limit + 1})",
                         custom_time_limit=limit,
                     )
                 ).rows[0][0]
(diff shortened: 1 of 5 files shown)


simonw:
This looks good, thanks.

I confirmed that the current Datasette 0.65 branch returns `[[foo]]` but the Datasette 1.0 alphas return `"[foo]"` like this:
```
/tmp % uv run --with datasette==0.65 python \
  -c 'from datasette.utils import escape_sqlite; print(escape_sqlite("[foo]"))'
Installed 28 packages in 21ms
[[foo]]
/tmp % uv run --with datasette==1.0a36 python \
  -c 'from datasette.utils import escape_sqlite; print(escape_sqlite("[foo]"))'
Installed 32 packages in 23ms
"[foo]"
```

PR #2857, opened 2026-07-25
Upgrade to ruff>=0.16.0

simonw:
See https://astral.sh/blog/ruff-v0.16.0

### DIFF
--- a/datasette/_pytest_plugin.py
+++ b/datasette/_pytest_plugin.py
@@ -89,7 +89,8 @@ def pytest_runtest_protocol(item, nextitem):
                 continue
             try:
                 ds.close()
-            except Exception as e:
+            except Exception as e:  # noqa: BLE001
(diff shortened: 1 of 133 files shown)


simonw:
After upgrading the version in `pyproject.toml`:
```
uv run ruff check .
```
Produces:
```
Found 862 errors.
[*] 633 fixable with the `--fix` option (60 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
Full output here: https://gist.github.com/simonw/3935f0e8042d117427ccdb5d4c3a13bf

simonw:
Running this command dropped it down to 154 errors:
```bash
uv run ruff check . --fix --unsafe-fixes
```

simonw:
In Claude Code with Opus 5 I prompted:
```
Run this command, then fix all the problems

    uv run ruff check .
```

PR #2871, opened 2026-08-05
Fix transaction wrapping for sqlite-utils 4.0 compatibility

### LINKED ISSUE #2831: Update transactions code for sqlite-utils 4.0
The 1.0a36 alpha upgraded to `sqlite-utils` 4.0 which introduced some surprising new reaction bugs.

I ran this against Fable 5:

> Clone simonw/datasette and simonw/sqlite-utils
> 
> Only consider the current main branches of each - construct a report on how Datasette uses database transactions at the moment, documenting the way it splits responsibility with sqlite
> 
> Include a section at the end with recommendations for making this more performant and robust

Report here: https://gist.github.com/simonw/539ef461154ee0f5938f060a8475532d

A bunch of stuff in there, here's the most important:

> **The `transaction=True` wrapper is inert for sqlite-utils-based writes.** `with conn:` opens no transaction at entry (verified: `conn.in_transaction` is `False` inside the block). So when a write function's first statement runs through sqlite-utils, sqlite-utils finds no open transaction, opens and *commits its own*, and Datasette's commit/rollback at task end operates on nothing. Empirically (appendix): a task that runs `insert_all()` of five rows and then raises leaves all five rows committed and visible to a concurrent reader *before* the task even fails; a task that does two sqlite-utils inserts where the second hits an `IntegrityError` leaves the first committed.

bunlongheng:
## Problem

sqlite-utils 4.0 changed transaction behaviour: every write (`insert_all`, `upsert_all`, `execute`, etc.) now checks `conn.in_transaction` at call-time. If that flag is `False`, sqlite-utils opens and immediately commits its **own** transaction, bypassing Datasette's outer `BEGIN IMMEDIATE`. A subsequent failure leaves the committed rows behind with no rollback path.

The old Datasette pattern:

```python
with conn:
    conn.execute("BEGIN IMMEDIATE")
    result = fn(conn)
```

relies on Python's `sqlite3.Connection` context manager calling `conn.commit()` on exit. This is fragile: it does not guarantee that `conn.in_transaction` is `True` during `fn()`, and if an inner `with conn:` inside the write function already committed, the outer `with conn:` exit becomes a no-op commit on a closed transaction.

## Fix

Replace both the non-threaded path (`database.py:352`) and the write-thread path (`database.py:491`) with an explicit `try/except`:

```python
conn.execute("BEGIN IMMEDIATE")
try:
    result = fn(conn)
    conn.commit()
except BaseException:
    conn.rollback()
    raise
```

Key properties:
- `conn.in_transaction` is `True` from `BEGIN IMMEDIATE` until `commit()`/`rollback()`, so sqlite-utils 4.0 correctly detects the outer transaction and uses `SAVEPOINT` instead of auto-commit
- Python's `conn.commit()` / `conn.rollback()` are idempotent no-ops when no transaction is active, so write functions that manage their own inner `with conn:` (e.g. `alter_table`) continue to work

## Testing

- All 147 `test_api_write.py` tests pass
- All 67 `test_internals_database.py` tests pass

Closes #2831

### DIFF
--- a/datasette/database.py
+++ b/datasette/database.py
@@ -349,9 +349,19 @@ class Database:
                 self._write_connection = self.connect(write=True)
                 self.ds._prepare_connection(self._write_connection, self.name)
             if transaction:
-                with self._write_connection:
-                    self._write_connection.execute("BEGIN IMMEDIATE")
+                # Use explicit BEGIN IMMEDIATE so that conn.in_transaction is
+                # True during fn(), allowing sqlite-utils 4.0 to detect the
+                # outer transaction and use savepoints instead of auto-committing
+                # each individual write. Use conn.commit()/conn.rollback() rather
+                # than conn.execute("COMMIT"/"ROLLBACK") so that calls are no-ops
+                # when the inner function has already committed or rolled back.
+                self._write_connection.execute("BEGIN IMMEDIATE")
+                try:
                     result = fn(self._write_connection)
+                    self._write_connection.commit()
+                except BaseException:
+                    self._write_connection.rollback()
+                    raise
             else:
                 result = fn(self._write_connection)
         else:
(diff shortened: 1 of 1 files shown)


simonw:
Thanks for highlighting this.

In adding a test for this I found the following bug, which this PR doesn't fix. I'm going to work on that separately from this.
- #2924

PR #2873, opened 2026-08-08
Keyboard accessibility and ARIA for column cog menu

opensource-joe:
Refs #1964

The column actions (cog) menu is currently mouse-only. The trigger is a bare SVG with a click handler — it isn't in the tab order — and the menu has no ARIA roles or keyboard navigation.

This adds keyboard support following the [ARIA menu button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/):

- The cog trigger becomes a focusable `role="button"` with `aria-haspopup="menu"`, `aria-expanded`, and an `aria-label` ("Column actions for {column}"). It opens with **Enter / Space / ArrowDown**.
- The menu gets `role="menu"` and each item `role="menuitem"` with roving `tabindex`; **Up/Down/Home/End** move focus (wrapping).
- **Escape** closes and returns focus to the trigger; **Tab** closes the menu.

Mouse behavior is unchanged. Diff is limited to `datasette/static/table.js` (+71/-2).

Note: I wasn't able to run the full Playwright suite headless in my environment, so a maintainer check in a browser would be welcome — happy to add a Playwright test for the keyboard flow if you'd like.

### DIFF
--- a/datasette/static/table.js
+++ b/datasette/static/table.js
@@ -500,6 +500,8 @@ function renderActionLink(itemConfig) {
   var newLink = document.createElement("a");
   newLink.textContent = itemConfig.label;
   newLink.href = itemConfig.href || "#";
+  newLink.setAttribute("role", "menuitem");
+  newLink.tabIndex = -1;
   if (itemConfig.onClick) {
     newLink.addEventListener("click", itemConfig.onClick);
   }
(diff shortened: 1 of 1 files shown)


opensource-joe:
@simonw Would you be able to approve the workflow runs on this PR?

As a first-time contributor here my workflows need a maintainer to release them, so all six runs are sitting at `action_required` and none have executed since the PR was opened on 8 August. It is easy to miss because the PR reads as green at a glance: `docs/readthedocs.org` reports and passes, so the checks list looks healthy while the actual test suite has never run.

Not asking for a review yet, just for CI to be allowed to run so anything it catches is mine to fix first.

PR #2886, opened 2026-08-31
Run datasette serve startup and uvicorn on a single event loop

asg017:
<details>

First PR in a stack that reworks Datasette's startup/shutdown lifecycle around the ASGI lifespan protocol.

## What this does

`datasette serve` currently runs `check_databases()` and `invoke_startup()` via `run_sync()` on a temporary event loop, closes that loop, and then `uvicorn.run()` opens a brand new one to serve requests. Anything a plugin's `startup` hook attaches to the running loop — `asyncio.create_task(...)` background tasks, `Lock`/`Queue`/`Event` objects — is orphaned when the temporary loop is torn down. datasette-litestream's startup hook, which schedules a background task with `asyncio.get_running_loop().create_task(...)`, is the motivating real-world case.

This PR wraps the serve path in a single `_serve_async()` coroutine driven by one `asyncio.run()`: soundness checks, startup hooks, and `uvicorn.Server(...).serve()` all share the same event loop.

## Changes

- `datasette/cli.py`: the `--get` path keeps its own `run_sync()` startup calls; the real serve path moves startup into `_serve_async()` and switches from `uvicorn.run()` to `uvicorn.Server(uvicorn.Config(...)).serve()` awaited on the shared loop.
- `pyproject.toml`: uvicorn minimum bumped from `>=0.11` to `>=0.29` for the `Server`/`Config` API used here.
- `tests/conftest.py`: new `serve_with_plugins` factory fixture that starts `datasette serve` in a subprocess with plugins written to a temporary `--plugins-dir`. These tests need the real serve path (event-loop wiring, exit codes, signals), which the in-process `pm.register` plugin pattern can't reach.
- `tests/test_cli_serve_server.py`:
  - `test_startup_hook_background_task_runs_on_serving_loop`: litestream-shaped regression test; fails on unmodified `main`.
  - `test_startup_error_fails_fast_before_port_binds`: characterization test pinning that a `StartupError` still exits non-zero with the message printed, before the port binds. It guards the later PRs in this stack that move startup into the ASGI lifespan.

## Notes

- Behavior of `--get` and the exit paths is unchanged.
- Later PRs in the stack build on this: startup via ASGI lifespan, a supervised background-tasks API, a `shutdown()` plugin hook, and docs.

</details>


simonw:
#2886, #2887, #2889 ready to review today.

simonw:
GPT-5.6 Sol identified an interesting issue:

> Calling `asyncio.run()` directly bypasses Uvicorn’s event-loop selection. When `uvloop` is installed, Datasette now silently uses the standard `asyncio` loop, potentially reducing performance and ignoring Uvicorn loop configuration. The outer loop should use the factory selected by `uvicorn.Config`, with compatibility for supported Uvicorn versions.

I don't think we should fix that here though, since by default Datasette uses Uvicorn without `uvloop`. I'll open a separate issue for later to investigate `uvloop` optimizations.

datasette/cli.py
```diff
@@ -663,23 +663,23 @@ def serve(
...
+        # Run async soundness checks before startup hooks, since invoke_startup
+        # now populates internal tables which requires querying each database
+        run_sync(lambda: check_databases(ds))
+
+        # Run the "startup" plugin hooks
+        try:
+            run_sync(ds.invoke_startup)
+        except StartupError as e:
+            raise click.ClickException(e.args[0])
+
```
simonw: I guess we still need this older style of running invoke startup for the `datasette .. --get /` command, which doesn't use Uvicorn at all.

PR #2887, opened 2026-08-31
Run startup via ASGI lifespan instead of waiting for the first request

asg017:
<details>

Second PR in the startup/lifecycle stack, on top of #2886.

## What this does

Moves Datasette startup from "lazily on the first HTTP request" to the ASGI lifespan protocol. A real deployment now finishes the immutable-database table-count precompute and all `startup` plugin hooks before the server accepts its first request; the first-request path remains only as a fallback for hosts that never send lifespan events (such as `httpx.ASGITransport`, which `datasette.client` uses).

## Changes

- `datasette/app.py`: new `Datasette._startup_sequence()` — an idempotent, lock-guarded single entry point that runs the immutable-db table-count precompute and then `invoke_startup()`. It tracks `_setup_db_done` separately from `_startup_invoked` so a caller that ran a bare `invoke_startup()` first (as `datasette serve` does) cannot cause the table-count precompute to be skipped. `AsgiLifespan` now gets `on_startup=[self._startup_sequence]`; `AsgiRunOnFirstRequest` calls the same method as fallback.
- `datasette/utils/asgi.py`:
  - `AsgiLifespan` catches startup/shutdown exceptions and reports them as `lifespan.startup.failed` / `lifespan.shutdown.failed` messages instead of letting them escape the protocol.
  - `AsgiRunOnFirstRequest` gains an `asyncio.Lock` with a double-checked `_started` flag, so concurrent first requests cannot interleave with startup; it also now ignores `lifespan`-type scopes so lifespan errors are handled by `AsgiLifespan`'s try/except.
- `datasette/cli.py`: `_serve_async()` calls `ds._startup_sequence()` instead of bare `invoke_startup()`, sharing the idempotent entry point with the lifespan event uvicorn sends moments later.

## Behavior notes

- `AsgiRunOnFirstRequest` now marks itself started only after the hooks succeed, so a failed startup is retried on the next request instead of being poisoned after one attempt.
- Startup hooks run at most once regardless of which paths fire (lifespan, first request, CLI) — covered by tests.

## Tests

New `tests/test_lifespan.py` with seven tests: hand-driven `lifespan.startup` with no HTTP request, startup failure reporting, exactly-once semantics across lifespan + first request, the no-lifespan fallback via raw `ASGITransport` and via `datasette.client`, ten concurrent first requests against a slow startup hook, and a regression test that table counts still precompute when `invoke_startup()` ran first.

</details>

### DIFF
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -2815,15 +2815,17 @@ class Datasette:
         This is the single entry point used by both AsgiLifespan (so
         real deployments finish startup before accepting requests) and
         AsgiRunOnFirstRequest (the fallback for hosts that never send
-        lifespan events, e.g. DatasetteClient's httpx.ASGITransport). It
-        may also race an explicit `await ds.invoke_startup()` call made by
-        `datasette serve` before the server starts serving - that's fine,
-        `invoke_startup()` has its own `_startup_invoked` guard.
+        lifespan events, e.g. DatasetteClient's httpx.ASGITransport), and
+        `datasette serve` (cli.py) calls it too. The fast path below checks
+        both `_startup_invoked` and `_setup_db_done` - not just the former -
+        so that a bare `await ds.invoke_startup()` made by a caller ahead of
+        `_startup_sequence()` (which only sets `_startup_invoked`) can't
+        make this method skip the immutable-database table-count precompute.
         """
-        if self._startup_invoked:
+        if self._startup_invoked and self._setup_db_done:
             return
         async with self._startup_lock:
-            if self._startup_invoked:
+            if self._startup_invoked and self._setup_db_done:
                 return
             if not self._setup_db_done:
                 # First time server starts up, calculate table counts for
--- a/datasette/app.py
+++ b/datasette/app.py
@@ -453,8 +453,10 @@ class Datasette:
         self.databases = collections.OrderedDict()
         self.actions = {}  # .invoke_startup() will populate this
         self._column_types = {}  # .invoke_startup() will populate this
+        self._setup_db_done = False
         try:
             self._refresh_schemas_lock = asyncio.Lock()
+            self._startup_lock = asyncio.Lock()
         except RuntimeError as rex:
             # Workaround for intermittent test failure, see:
             # https://github.com/simonw/datasette/issues/1802
@@ -462,6 +464,7 @@ class Datasette:
                 loop = asyncio.new_event_loop()
                 asyncio.set_event_loop(loop)
                 self._refresh_schemas_lock = asyncio.Lock()
+                self._startup_lock = asyncio.Lock()
             else:
                 raise
         self.crossdb = crossdb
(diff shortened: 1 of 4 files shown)


datasette/utils/asgi.py
```diff
@@ -624,10 +637,23 @@ def __init__(self, asgi, on_startup):
...
-            self._started = True
-            for hook in self.on_startup:
-                await hook()
+        # Leave "lifespan" scope events alone - this shim only exists as a
+        # fallback for hosts that never send them. It wraps AsgiLifespan, so
+        # if it ran on_startup here too, a startup exception would escape
+        # before AsgiLifespan's own try/except got a chance to turn it into
+        # a lifespan.startup.failed message.
+        if scope["type"] != "lifespan" and not self._started:
+            async with self._lock:
```
simonw, line 652: This lock seems smart.

datasette/app.py
```diff
@@ -2803,24 +2806,52 @@ async def resolve_row(self, request):
...
         if self.setting("trace_debug"):
             asgi = AsgiTracer(asgi)
-        asgi = AsgiLifespan(asgi, on_shutdown=[_close_on_shutdown])
-        asgi = AsgiRunOnFirstRequest(asgi, on_startup=[setup_db, self.invoke_startup])
+        asgi = AsgiLifespan(
+            asgi,
+            on_startup=[self._startup_sequence],
+            on_shutdown=[_close_on_shutdown],
+        )
+        asgi = AsgiRunOnFirstRequest(asgi, on_startup=[self._startup_sequence])
```
simonw, line 2854: I like that we ditched the `[setup_db, self.invoke_startup()]` list here for a combined `self._startup_sequence`.

simonw:
Feedback from GPT-5.6 Sol (which I didn't think was worth acting on)

> - **Medium:** Immutable databases added by `startup()` hooks miss table-count precomputation. `_setup_db_done` is set before hooks execute, so newly added immutable databases never receive the long-timeout count pass. Large databases can consequently show missing counts.
> - **Low:** Lifespan startup and shutdown failures are converted to strings without logging their traceback. Uvicorn reports the error message but no source location, making plugin failures difficult to diagnose.

<details><summary>More details</summary>

Validation included the full post-merge `main` diff, focused lifecycle/startup/CLI/plugin tests, direct reproduction of both issues, and `git diff --check`. No tracked files were changed.

::code-comment{title="[P2] Precompute databases added during startup" body="The setup-complete flag is set before startup hooks execute. If a documented startup hook adds an immutable database, later lifecycle calls take the fast path and that database never receives the long-timeout table-count pass. Run precomputation after hooks complete or track completion per database." file="/Users/simon/.codex/worktrees/37f4/datasette/datasette/app.py" start=2836 priority=2}

::code-comment{title="[P3] Preserve lifespan failure tracebacks" body="Converting every exception to str(e) means Uvicorn logs only the message, with no traceback or source location. Log the caught exception with exc_info before sending the lifespan failure response; apply the same treatment to the shutdown handler." file="/Users/simon/.codex/worktrees/37f4/datasette/datasette/utils/asgi.py" start=307 priority=3}

</details>

PR #2889, opened 2026-08-31
Add datasette.add_background_task() with supervised launch after startup

asg017:
This PR adds a new `datasette.add_background_task()` API for plugins. Plugins like cron, litestream, and dozens of others often need to run code "in the background", ie not during an HTTP request or on startup. This is often stuff like montoring code, backups, etc.

```py
import httpx2

async def heartbeat(datasette):
    while True:
        httpx.get("https://hc-ping.com/c8ea6586a5e7a")
        await asyncio.sleep(60)


@hookimpl
def startup(datasette):
    datasette.add_background_task(heartbeat, name="heartbeat")
```

This was possible before, but required some weird ASGI workarounds and only started after the first HTTP request. But now that #2887 is in, this works as expected. 

<details>

Third PR in the startup/lifecycle stack, on top of #2887.

## What this does

Adds `datasette.add_background_task(func, name=None)` — supervised background work for plugins, replacing fire-and-forget `asyncio.create_task()` calls in `startup` hooks. Core owns the task references (no silent garbage collection), the launch timing, crash surfacing, and cancellation.

## Changes

- New `datasette/background_tasks.py`: `BackgroundTask` handle (states `registered` → `running` → `completed`/`crashed`/`cancelled`, with `.exception`, `.started_at`, `.plugin`, `.cancel()`) and `BackgroundTaskSupervisor` (strong references for the process lifetime, idempotent `launch_all()`, `cancel_all()` with a grace period, crash logging with full traceback to the `datasette.background_tasks` logger).
- `datasette/app.py`: `add_background_task()` buffers registrations until launch, or starts immediately if launch already happened; `start_background_tasks()` is a public entry point for embedders and headless CLI uses; private `_launch_background_tasks()` is wired as the second `on_startup` entry in both `AsgiLifespan` and `AsgiRunOnFirstRequest` — after `_startup_sequence`, so launch happens only once every plugin's `startup` hook has had a chance to register work.
- `datasette/cli.py`: `--get` suppresses background-task launch — its one-shot in-process request flows through the full ASGI stack but must never start long-lived work.

## Docs

Rolled down from the former docs-only tip PR (#2893) so the API lands documented:

- `docs/internals.rst`: full reference for `add_background_task()`, `start_background_tasks()` and `BackgroundTask` objects, plus a new "Application lifecycle" section (`datasette_lifecycle`) describing the startup → launch → serving → shutdown sequence and the three startup trigger paths. The lifecycle text covers only what exists at this point in the stack; the shutdown hook, wrapper-timing guarantee and `/-/tasks` cross-references are woven in by the later PRs that introduce those features.
- `docs/plugin_hooks.rst`: `startup()` now documents the same-event-loop guarantee and the background-task use-case.
- `docs/testing_plugins.rst`: how to launch registered tasks in tests via `start_background_tasks()`.

## Notes

- Completed/crashed task handles are deliberately never pruned in this version (full introspection for the `/-/tasks` endpoint later in the stack); a pruning policy can come later if unbounded growth shows up in practice.
- Task names default to `func.__qualname__` with `-2`, `-3`… suffixes on collision.

## Tests

New `tests/test_background_tasks.py` (10 tests): launch ordering relative to startup hooks, exactly-once launch across concurrent first requests, post-launch immediate start, cancellation semantics including grace-period stragglers, crash logging, name collisions, and the bare-`Datasette` embedder path. Plus a `tests/test_cli_serve_get.py` sentinel test that `--get` registers but never launches tasks.

</details>


datasette/app.py
```diff
@@ -2836,6 +2839,63 @@ async def _startup_sequence(self):
                 self._setup_db_done = True
             await self.invoke_startup()
 
+    def add_background_task(self, func, name=None) -> BackgroundTask:
```
simonw, line 2842: I was going to complain about lacking documentation, but that's in a later PR: https://github.com/simonw/datasette/pull/2893/changes#diff-a94fc1e6fae9a7b965684e35350146310ade0b63e5993f513e3a6f1f78056494R1436

simonw:
I had Codex generate its own docs for the new methods to help me review them: https://gist.github.com/simonw/10f35afc79e3ecb12d0d0c6f80cf873a

It wrote this example:

```python
async def refresh_cache(datasette):
    while True:
        # Refresh the cache here
        await asyncio.sleep(60)


@hookimpl
def startup(datasette):
    datasette.add_background_task(
        refresh_cache,
        name="refresh-cache",
    )
```

tests/test_background_tasks.py
```diff
@@ -0,0 +1,379 @@
...
+@pytest.mark.asyncio
+async def test_tasks_registered_in_startup_hook_run_after_lifespan_startup():
+    # Two tasks registered by one plugin's startup hook - order preserved,
+    # both running after lifespan startup completes, and no HTTP request
+    # of any kind is issued anywhere in this test.
+    events = []
+
+    async def task_one(datasette):
+        events.append("task_one")
+        await asyncio.Event().wait()
```
simonw, line 63: This is necessary so that the later assertion `assert handles[0].state == "running"` works - without the `wait()` the task would have ended.

datasette/background_tasks.py
```diff
@@ -0,0 +1,254 @@
...
+            if not plugin_module_name:
+                continue
+            if module_name == plugin_module_name or module_name.startswith(
+                plugin_module_name + "."
+            ):
+                return pm.get_name(plugin)
+    except Exception:  # noqa: BLE001
+        # Never let plugin-name resolution break task registration.
+        return None
+    return None
```
simonw, line 82: I think we should ditch this bit. It exists purely so that `/-/tasks` can show us the name of the plugin, but it's only guessing from the filepath - so we could instead show the filepath and let the user who is viewing that page guess themselves.

I'll file a separate issue for this so it doesn't block the rest of the reviews.

docs/internals.rst
```diff
@@ -1403,7 +1403,139 @@ Release all resources held by this ``Datasette`` instance. This calls :ref:`data
...
+    @hookimpl
+    def startup(datasette):
+        datasette.add_background_task(
+            poll_for_updates, name="my-plugin-poller"
+        )
+
+Core owns the task for the rest of the process's life:
+
+- **A strong reference is kept forever**, so the task can never be silently garbage collected the way an unreferenced ``asyncio.create_task()`` call can be.
+- **A crash is logged, not swallowed.** If ``func`` raises anything other than ``asyncio.CancelledError``, the exception (with its traceback) is logged to the ``datasette.background_tasks`` logger and recorded on the handle's ``.exception``, and the task's ``.state`` becomes ``crashed``. **There is no automatic restart in v1** — a long-running loop should catch and log its own transient errors internally if it wants to keep running after one.
```
simonw: > **There is no automatic restart in v1** — a long-running loop should catch and log its own transient errors internally if it wants to keep running after one.

We should drop that, v1 doesn't mean anything here.

simonw:
I had GPT-6 Astra build a demo plugin to help me try this out: https://gist.github.com/simonw/e8fa7adfc7f85a77c87116f6f601ed3c

CleanShot 2026-09-15 at 10 56 40 AM@2x

PR #2890, opened 2026-08-31
Add a shutdown() plugin hook with ordered graceful teardown

asg017:
This PR adds a new `shutdown()` hook to compliment the `startup()` hook we have. Oftentimes a plugin will needs some signal when the server is shutting down, either for cleanup or to safely flush out any queues/buffers, like with [`datasette-litestream`](https://github.com/datasette/datasette-litestream) (which needs to safely shutdowna daemon at exittime). 

Here's a sample usage, where a plugin logs Datasette Events to the internal database. Writes are buffered to avoid starving the write thread. But if a shutdown happens and there's still items in the buffer, the plugin can notice in the new `shutdown()` hook and properly cleanup before exiting. 

```python
import asyncio
import json


class EventBuffer:
    def __init__(self):
        self.pending = []
        self.stop = asyncio.Event()

    async def flush(self, datasette):
        if not self.pending:
            return
        batch, self.pending = self.pending, []
        db = datasette.get_internal_database()
        await db.execute_write_many(
            "insert into event_log (event, properties) values (?, ?)",
            batch,
        )

    async def run(self, datasette):
        while not self.stop.is_set():
            try:
                await asyncio.wait_for(self.stop.wait(), timeout=10)
            except asyncio.TimeoutError:
                pass
            await self.flush(datasette)


buffer = EventBuffer()


@hookimpl
def startup(datasette):
    async def inner():
        await datasette.get_internal_database().execute_write(
            "create table if not exists event_log (event text, properties text)"
        )
        datasette.add_background_task(
            lambda: buffer.run(datasette), name="event-log-flusher"
        )

    return inner


@hookimpl
def track_event(datasette, event):
    buffer.pending.append((event.name, json.dumps(event.properties())))


@hookimpl
def shutdown(datasette):
    async def inner():
        buffer.stop.set()
        await buffer.flush(datasette)

    return inner
```

One thing to keep in mind, the `shutdown()` hook fires *before* background tasks are cancelled, to give plugins a chance to clean them up themselves. Not entirely sure if this is the correct way to do it, but seems to be the most flexible to me. 

<details>

Fourth PR in the startup/lifecycle stack, on top of #2889.

## What this does

Adds a `shutdown(datasette)` plugin hook and an ordered graceful-teardown sequence, wired to the ASGI `lifespan.shutdown` event — so Ctrl-C or SIGTERM on `datasette serve` gives plugins a chance to clean up.

## Changes

- `datasette/hookspecs.py`: new `shutdown(datasette)` hookspec; like `startup()` it can be sync or return an async function.
- `datasette/app.py`: new `Datasette.invoke_shutdown()`, replacing the old inline `_close_on_shutdown` as the `AsgiLifespan` shutdown handler. It is idempotent (a second `lifespan.shutdown` from a misbehaving host won't re-run teardown) and runs in a deliberate order:
  1. plugin `shutdown` hooks — while background tasks are still alive, so a plugin can coordinate with its own task (e.g. tell a queue consumer to stop pulling work) before cancellation;
  2. `cancel_all()` on supervised background tasks, with a 5s grace period;
  3. `close()` — last, so hooks and cancelled tasks still had working database connections.

  A hook that raises is logged and swallowed so one broken plugin can't skip another plugin's cleanup or the rest of teardown.
- `docs/plugin_hooks.rst`: documents the hook, including the guarantee that it runs before task cancellation and connection close, with cross-references into the lifecycle and background-task docs (rolled down from the former docs-only tip PR #2893).
- `docs/internals.rst`: the "Application lifecycle" section's shutdown step now describes shutdown-hook ordering (also rolled down from #2893).

## Tests

- New `tests/test_shutdown.py`: hook invocation via hand-driven lifespan messages, ordering guarantees, idempotency, and exception swallowing.
- `tests/test_cli_serve_server.py`: end-to-end subprocess tests that SIGTERM and SIGINT on a real `datasette serve` process run a plugin's `shutdown` hook (checked via a sentinel file written by the hook) — proving the PR #2886 serve path actually delivers signals through to uvicorn's `lifespan.shutdown`.

</details>


asg017:
Found a gap while running datasette-paper on this stack: the `shutdown()` hook fires too late for plugins that hold **long-lived streaming responses** (SSE, long-poll).

**Repro:** a plugin route that streams SSE indefinitely (paper's `/events`), open it in 3 browser tabs, Ctrl-C `datasette serve`. The server logs `Waiting for connections to close. (CTRL+C to force quit)` and hangs until every tab is closed. Repeated Ctrl-C doesn't force quit either.

**Why:** `invoke_shutdown()` hangs off ASGI `lifespan.shutdown`, and uvicorn (0.46) only sends that at the very end of `Server.shutdown()`:

```python
# uvicorn/server.py
for connection in list(self.server_state.connections):
    connection.shutdown()          # in-flight response: only sets keep_alive = False
await asyncio.wait_for(self._wait_tasks_to_complete(), timeout=self.config.timeout_graceful_shutdown)  # None by default
if not self.force_exit:
    await self.lifespan.shutdown() # <- shutdown() hooks run here
```

1. **Graceful path:** `_wait_tasks_to_complete` waits for every connection to close. A streaming handler never learns shutdown has started (no `http.disconnect` is sent), so it never ends, so the `shutdown()` hook, the one place a plugin could tell it to stop, never runs. `cancel_all(grace=5.0)` doesn't help either: request handlers aren't supervised background tasks, and it runs after the drain.
2. **Force path (second Ctrl-C):** `force_exit` breaks the wait loop but skips `lifespan.shutdown()` entirely, and `_wait_tasks_to_complete` still ends with `await server.wait_closed()`. On Python ≥ 3.12.1, `asyncio.Server.wait_closed()` waits until all connections have dropped, so force-quit blocks on the same sockets.

`test_sigterm_runs_shutdown_hooks` / `test_sigint_runs_shutdown_hooks` pass because nothing holds a connection open when the signal arrives. A variant with one open streaming request would reproduce the hang.

**Possible fix:** give plugins a signal when shutdown *begins*, before the drain. `cli.py` already constructs the `uvicorn.Server`, so e.g. subclass it and override `shutdown()` to set a `datasette.shutting_down` `asyncio.Event` (or fire a separate `shutdown_started` hook) before calling `super().shutdown()`. Streaming handlers await that alongside their disconnect watcher and end their responses; connections close, the drain and `wait_closed()` complete, and the existing `shutdown()` hook runs as today. Passing `timeout_graceful_shutdown` would be a sensible backstop for handlers that ignore it.

Caveat: only covers `datasette serve`. ASGI has no "shutdown starting" message, so under gunicorn/bare uvicorn the event would never be set and plugins would still need a fallback.

Will follow up here.

simonw:
Annoying bug in stacked PRs: if you hit "squash and merge" and edit the commit message it ignores your edits. I've had to force-push some modified commits to main instead.

PR #2891, opened 2026-08-31
Run plugin asgi_wrapper middleware inside the startup-arming layer

asg017:
Some ASGI middlware fixes. To be honest, not 100% sure how this is relevant or what it fixes, but Claude was insistent. Seems to handle some corner cases of plugin hooks that need to be ran after startup finishes?

<details>

Fifth PR in the startup/lifecycle stack, on top of #2890.

## What this does

Reorders the ASGI middleware assembly in `Datasette.app()` so plugin `asgi_wrapper` middleware runs **inside** the startup-arming layer (`AsgiRunOnFirstRequest`) but still outside `AsgiLifespan`. Wrappers get a single, simple contract: every `http`/`websocket` scope they see arrives after startup has fully completed — on both the lifespan path and the first-request fallback.

## Why

Previously the wrapper loop was outermost, which had two consequences:

- a wrapper could see request scopes before startup ran (e.g. before another plugin's `startup` hook had run internal-database migrations);
- a short-circuiting wrapper — an auth plugin returning 401/403, a CORS plugin answering a preflight — that never called the inner app deferred startup indefinitely for those requests.

`lifespan` scopes are unaffected: `AsgiRunOnFirstRequest` ignores them, so plugins that inspect or wrap lifespan events see identical message flow.

## Changes

- `datasette/app.py`: the `pm.hook.asgi_wrapper` loop moves below `AsgiRunOnFirstRequest` in the stack; comment documents the layering contract.
- `docs/internals.rst`: the "Application lifecycle" section now states that startup completes before plugin `asgi_wrapper` middleware sees any request (rolled down from the former docs-only tip PR #2893). The wrapper-reorder release notes land in the changelog in #2892 at the top of the stack.

## Tests

Three new tests in `tests/test_lifespan.py`: a wrapper asserting startup has completed before any `http` scope (fails on the pre-reorder `app()`), a short-circuiting 403 wrapper that no longer prevents startup, and a recording wrapper pinning that `lifespan` scopes still reach wrappers.

</details>


simonw:
OK, i think I get it. Prior to this fix it was possible for this to happen:

**First request -> run plugin middleware -> triggers first run of startup -> process request**

That's a problem if the plugin middleware needed any tables that would have been created by a startup hook.

After this:

**First request -> triggers startup if needed -> run plugin middleware -> process request**

PR #2894, opened 2026-08-31
OpenTelemetry demo: pure-Python OTLP receiver and no-Docker Jaeger recipe

asg017:
<details>

Stacked on #2864. Adds `demos/otel/` — a runnable demonstration that Datasette's OpenTelemetry spans survive a real OTLP export, with no Docker, no collector, and nothing beyond `uv` (plus optionally the Jaeger binary for a UI).

## What's in it

- **`otlp_receiver.py`** (~150 lines, stdlib + `opentelemetry-proto`): an OTLP/HTTP receiver that decodes the actual protobuf wire format. It prints live batch counts as spans arrive, and on Ctrl-C prints a summary: per-span-name counts and total durations, the five slowest spans (annotated with truncated SQL text or HTTP status), and the root-span breakdown. A browser GET on the port explains that OTLP ingestion has no UI, rather than showing a bare 501.
- **`Justfile`**: `just receiver` **or** `just jaeger` for terminal 1 — both listen for OTLP/HTTP on 4318, so terminal 2 is one identical `just serve` either way. `just serve` generates a 200-row `demo.db` on first run and starts Datasette under `opentelemetry-instrument` with the right env vars; `just request` makes a traced request.
- **`README.md`**: both flows with real captured output, notes on the non-obvious env vars (`OTEL_BSP_SCHEDULE_DELAY=1000`, `OTEL_SERVICE_NAME`, metrics/logs exporters off), and the privacy story (SQL text is recorded, parameter values never).

## Verified

Both paths were exercised for real against this branch:

- Receiver: startup plus one request against the demo table exported **93 spans with exactly two roots** — the `GET <route>` request span and `datasette.startup`.
- Jaeger (v2 binary, no Docker): the same run lands as two traces — 67 spans nested under the request span (including `db.query` → `db.query.execute` nesting across thread boundaries) and 26 under startup.

## Compared to the earlier demo iteration

This is a deliberately trimmed port of the demo that lived on the exploratory `otel-6-docs` branch. Two of that demo's workarounds are now core behavior in this stack, so they are gone rather than ported:

- No `asgi_wrapper` plugin: the per-request `SpanKind.SERVER` span is emitted by core (#2863), so the old "75 root spans per request" problem no longer exists.
- No "ignore the ~24 tiny startup traces" caveat: startup spans nest under `datasette.startup`.

The in-memory waterfall script (`trace_demo.py`) was deliberately left out to keep this demo-only and small.

One observation from testing, left as-is: the request span is named after the matched route, which in Datasette is a regex (`GET /(?P<database>[^\/\.]+)/...`). Honest low-cardinality naming, but it reads poorly in a trace UI; if we want prettier names that's a change to the request-span PR, not this demo. The README calls it out.

</details>

### DIFF
--- a/demos/otel/Justfile
+++ b/demos/otel/Justfile
@@ -1,0 +1,30 @@
+# OpenTelemetry demo - see README.md in this directory.
+#
+#   just receiver  +  just serve     -> span summary in the terminal
+#   just jaeger    +  just serve     -> real trace UI at http://localhost:16686
+#
+# Both listen for OTLP/HTTP on port 4318, so `just serve` works with either
+# (but not both at once).
+
+otel_env := "OTEL_TRACES_EXPORTER=otlp OTEL_METRICS_EXPORTER=none OTEL_LOGS_EXPORTER=none OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 OTEL_SERVICE_NAME=datasette OTEL_BSP_SCHEDULE_DELAY=1000"
+
+default:
+    @just --list --unsorted
+
+# Terminal 1, option A: the ~150 line pure-Python receiver. Ctrl-C for a summary
+receiver:
+    uv run --with opentelemetry-proto python otlp_receiver.py
+
+# Terminal 1, option B: Jaeger from its own binary - no Docker. UI on :16686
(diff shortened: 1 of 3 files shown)


asg017:
Closing for now — dropping the demo directory from the upstream stack. #2897 has been rebased directly onto #2863 without the `demos/otel/` content.

PR #2903, opened 2026-09-10
Fix infinite loop when streaming CSV on SQL views

### LINKED ISSUE #2902: CSV streaming (?_stream=on) on a SQL view loops forever, re-serving the same page
In 1.0a38, streaming a **view** as CSV with `?_stream=on` never terminates. The stream advances one page, then repeats that same page indefinitely until `max_csv_mb` truncates the response. Streaming a **table** works correctly. The issue doesn't seem to exist in 0.65.3, the latest 0.x release.

## Context

We have recently deployed a new database served by Datasette 1.0a38, which exposes several SQL views as the public query surface. We first implemented CSV downloads with `?_size=max`, but some initial testing of the CSV streaming functionality suggested that a switch to `?_stream=on` could lighten the load on the server. That is when we found the conflict with views: streaming works on a table, but a single `?_stream=on` request against a view pins CPU generating hundreds of MB of duplicated rows on a small database and never completes.

## What happens

If you stream a 500-row view with `?_stream=on`, the response never ends. In a ~15s capture it produced roughly 1.19M data rows, but only **200 distinct primary-key values**. Datasette appears to emit page 1 (rows 1–100) once, then emit page 2 (rows 101–200) on every subsequent iteration, never advancing past it. Streaming the same data as a **table** returns all 500 rows exactly once.

The problem is likely due to the fact that tables and views paginate differently: tables paginate by keyset, views paginate by `OFFSET`, and only the `OFFSET` path is broken.

## Diagnosis and fix (proposed by Claude Opus after autonomous testing and debugging)

The streaming loop threads the pagination cursor back into `table_view_data` as a `_next` keyword argument. That argument is honored once, then discarded before the next cursor is computed.

In `datasette/views/table.py`, the kwarg is honored to build the page's `OFFSET`:

```python
# datasette/views/table.py:2054
_next = _next or request.args.get("_next")
```

Later in the same function, just before the next-page token is computed, `_next` is re-read from the request arguments and the honored kwarg is overwritten:

```python
# datasette/views/table.py:2229
_next = request.args.get("_next")
```

During a stream `_next` is never present in the request arguments. `stream_csv` passes it only as a kwarg, and rejects it as a query argument outright:

```python
# datasette/views/base.py:198
if request.args.get("_next"):
    raise BadRequest("_next not allowed for CSV streaming")
```

So line 2229 always sets `_next = None` in the streaming case. The next-token computation for a view then reads that `None`:

```python
# datasette/views/table.py:2425
if is_view:
    next_value = int(_next or 0) + page_size   # int(None or 0) + page_size == page_size, every time
else:
    next_value = path_from_row_pks(rows[-2], pks, use_rowid)  # derived from the last row, unaffected
```

`next_value` is therefore pinned at `page_size` for every iteration, so the loop re-requests `OFFSET page_size` forever. The table branch derives its token from the last row of the page, so it advances correctly and never loops.

Between line 2054 and line 2229 there is no other assignment to `_next`, so line 2229 is a redundant re-read that only discards the honored value. Removing it (or making it non-clobbering, `_next = _next or request.args.get("_next")`) lets the view branch see the real offset. With that single line changed, the same 500-row view streams 500 rows / 500 distinct keys and terminates.

## Not present in the 0.x line

0.65.3 has the equivalent honor line but no clobbering re-read before the token computation:

- `datasette/views/table.py:354` (0.65.3) honors `_next = _next or request.args.get("_next")`
- `datasette/views/table.py:627` (0.65.3) computes `next_value = int(_next or 0) + page_size` from the still-honored value

The clobbering re-read was introduced when the table view was reworked from the class-based `RowTableShared.data` method into the functional `table_view_data`. `main` still carries it at `table.py:2229`.

## Reproduction

```python
import sqlite3
con = sqlite3.connect("repro.db")
con.execute("CREATE TABLE base (id INTEGER PRIMARY KEY, name TEXT, val INTEGER)")
con.executemany(
    "INSERT INTO base (id, name, val) VALUES (?, ?, ?)",
    [(i, f"row-{i}", i * 7 % 1000) for i in range(1, 501)],
)
con.execute("CREATE VIEW v_base AS SELECT id, name, val FROM base")
con.commit()
con.close()
```

```
datasette repro.db --setting max_returned_rows 1000000
```

- `curl 'http://localhost:8001/repro/base.csv?_stream=on'` returns 500 rows, 500 distinct ids. Correct.
- `curl 'http://localhost:8001/repro/v_base.csv?_stream=on'` never completes; the body is rows 1–200 repeated until `max_csv_mb` cuts it off.

Count distinct ids in the captured body to see the loop:

```
curl -s --max-time 15 'http://localhost:8001/repro/v_base.csv?_stream=on' \
  | tail -n +2 | cut -d, -f1 | sort -u | wc -l   # -> 200, not 500
```

## Impact

- Any `?_stream=on` request against a view generates duplicated rows up to `max_csv_mb`, which is wasted CPU and memory per request and a broken export.
- Tables are unaffected.
- The fix is confined to the view (`OFFSET`) pagination path; the table (keyset) path already ignores the clobbered value, so removing the redundant re-read does not change table behavior.

## Versions

```
datasette: 1.0a38
sqlite:    3.53.1
python:    3.13
```

AnkitaAdvitot:
When streaming a SQL view as CSV with `?_stream=on`, Datasette was looping indefinitely after page 1, repeating page 2 over and over (fixes #2902).

### What happened
In `table_view_data`, the kwarg `_next` passed in by the streaming loop was being overwritten by `_next = request.args.get("_next")` at line 2229. Because streaming requests deliberately do not allow `_next` in query arguments, this set `_next` to `None`. When the view pagination computed the next offset, `int(_next or 0) + page_size` stayed stuck on `page_size` forever.

### The fix
- Removed the redundant line in `table.py` so the offset passed via kwargs is kept.
- Added an async test in `tests/test_csv.py` verifying that streaming a paginated view completes and returns all unique rows.


simonw:
Thanks for catching this.

Issue #2854, opened 2026-07-22
Columns not returned


cawoodm:
https://datasette.io/legislators/executive_terms.json?_extra=columns,count,primary_keys
Returns count and primary keys but not columns ❌

https://datasette.io/legislators/executive_terms.json?_extra=columns
Works fine ✅

simonw:
This was fixed in 03f1ffdf8fbf0ed7da46be48f3fcc3f4698e1e21

PR #2906, opened 2026-09-12
Test combined columns, count and primary_keys extras

RamiNoodle733:
Refs #2854.

The exact combination reported in #2854 (`?_extra=columns,count,primary_keys`) is not covered by the existing comma-separated extras test. Current `main` handles it correctly, so this adds a focused regression test asserting that all three requested extras are preserved together.

This protects the renderer behavior that previously allowed `columns` to disappear when extras were combined.

Tests were not run locally because this automation runtime cannot check out GitHub repositories over the network; the change is limited to the async table API regression test.


simonw:
We don't need an extra test here, the bug itself was fixed in 03f1ffdf8fbf0ed7da46be48f3fcc3f4698e1e21.

PR #2908, opened 2026-09-13
Support floating point values in numeric comparison filters

### LINKED ISSUE #1681: Numeric comparison filters fail for decimal and negative values in SQL

RamiNoodle733:
Closes #1681.

Numeric comparison filters currently coerce values only when `str.isdigit()` succeeds. That leaves values such as `0.1`, negative numbers, and scientific notation bound as text, which breaks comparisons against calculated numeric columns in SQLite views.

This changes numeric comparison coercion to try `int()` first and then `float()`, preserving non-numeric strings unchanged. It also adds regression coverage for integer, negative, floating-point and exponent values, plus the calculated-view case described in the issue.

Validation performed in this run: reproduced the SQLite behavior directly with an in-memory database, where binding `"0.1"` returns no rows for a calculated numeric view but binding `0.1` returns the expected rows. Repository CI is expected to run on this PR.


RamiNoodle733:
I found that my earlier #2876 already addresses #1681 and has already been reviewed, so this newer PR is a duplicate. Closing this one and keeping #2876 as the canonical contribution.

RamiNoodle733:
Closing as an accidental duplicate of #2876. I consolidated the stronger calculated-view regression coverage from this PR into #2876, which now also preserves non-finite inputs (`nan`, `inf`, `-inf`) instead of coercing them to special floats.

PR #2909, opened 2026-09-13
Fix float values in numeric comparison filters

### LINKED ISSUE #1681: Numeric comparison filters fail for decimal and negative values in SQL

RamiNoodle733:
Fixes #1681.

Numeric comparison filters currently convert parameters to integers only when `str.isdigit()` succeeds. That leaves decimal values such as `0.1` as strings, which breaks comparisons against calculated numeric columns in SQLite views as described in the issue. Negative values are affected for the same reason.

This updates numeric filter parameter conversion to try `int()` first, then `float()`, while leaving non-numeric strings unchanged. Existing integer behavior is preserved.

A focused regression case covers both `0.1` and `-2.5` in the existing filter unit tests.

Validation:
- Branch is based directly on current upstream `main` (`b338c6f5`).
- Diff is limited to `datasette/filters.py` and `tests/test_filters.py`.
- The conversion logic was exercised directly for integers, decimals, negatives, exponent notation, and non-numeric values.
- Full `pytest`/Black checks were not run locally because this automation environment does not provide a runnable repository checkout, so I am not claiming those checks passed.


RamiNoodle733:
Closing this as a duplicate of the earlier #2876, which already addresses the same #1681 numeric-filter coercion issue and has an existing review thread. Keeping a single PR avoids splitting review across equivalent changes.

RamiNoodle733:
Closing as an accidental duplicate of #2876. The canonical PR now includes the numeric coercion fix, calculated-view regression coverage, and handling that keeps non-finite inputs (`nan`, `inf`, `-inf`) as strings instead of binding special float values.

PR #2911, opened 2026-09-14
Slim down CI runs per commit

asg017:
Our CI/CD can be slow at times. In #2352 I added the "feature" where we ran the Datasette test suite in a matrix across multiple SQLite and Python versions, and tbh it's overkill. Since then those tests only really break if it's an old SQLite version across all Python version, so this PR updates that so we only test old SQLite once, instead of the matrix.

Also updated some other GHA workflows: concurrency checks to stop all jobs on failures to save time, test only on main and pull requests (would run on every push before, wasteful). Also moved some coverage tests around.

An agent estimates that will go from 44 jobs to 13, and ~160 minutes to 42. Not sure about those exact numbers, but should really save some time around here. 

<details>

Every commit on a branch pushed to this repo with an open PR currently triggers each workflow twice (`push` and `pull_request`), plus a 10-job SQLite matrix. That's about 44 jobs / ~161 job-minutes per commit, measured from recent runs.

### Changes

- **`push` only on `main`** for Test, Test SQLite versions, Playwright, Pyodide, spellcheck and Prettier. PR commits are covered by `pull_request`. Added `workflow_dispatch` for manual runs.
- **Test SQLite versions** is now a single job: Python 3.13 + SQLite 3.25 (was 5 Pythons × 2 SQLite versions).
- **Coverage folded into the Python 3.14 test job** (now includes serial tests via `--cov-append`); `test-coverage.yml` removed and `codecov-action` bumped from v1 to v5.
- **Concurrency groups** cancel superseded in-progress runs on PRs. Runs on `main` always complete.
- **Prettier** now also runs on pull requests (it was push-only).

The full Python matrix (3.10–3.15) and all three Playwright browsers are unchanged.

### Estimated impact per PR commit

| | Jobs | Job-min |
|---|---|---|
| Before | 44 | ~161 |
| After | 13 | ~42 |

### Notes

- If branch protection requires the "Calculate test coverage" check, that requirement needs removing.
- Branches without an open PR no longer get CI on push.

https://claude.ai/code/session_012G7fa8HqJy9AkSjkocetAz

</details>


.github/workflows/spellcheck.yml
```diff
@@ -1,6 +1,15 @@
 name: Check spelling in documentation
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - main
+  pull_request:
```
simonw, line 7: I like this pattern. Previously we ran every CI thing twice on a push to PR - once for the branch, once for the PR.

.github/workflows/test.yml
```diff
@@ -12,6 +21,9 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14", "3.15"]
+        include:
+        - python-version: "3.14"
+          coverage: true
```
simonw, line 26: Nice, this replaces our separate `test-coverage.yml`.

PR #2912, opened 2026-09-14
docs: fix dead Starlette, Uvicorn and Docker Hub links

auroraxo:
## Summary

Three documentation links are dead. I found them with a link-rot scan over `simonw/datasette@main` (31 docs files, 863 outbound links checked) and verified each one by hand before opening this PR.

**Two dead domains.** `www.starlette.io` and `www.uvicorn.org` no longer resolve at all — both return NXDOMAIN, not a redirect, so there is nothing for a reader to follow:

```
$ dig +short www.starlette.io    # NXDOMAIN (Status 3)
$ dig +short www.uvicorn.org     # SERVFAIL/NXDOMAIN (Status 2)
```

Both projects moved their documentation to new domains. Starlette's own repository homepage is now `https://starlette.dev`, and Uvicorn's docs are at `https://uvicorn.dev`.

**One malformed Docker Hub URL.** `https://hub.docker.com/datasetteproject/datasette` returns 404 — Docker Hub user/org repositories live under `/r/`.

## Changes

| File | Old | New |
|---|---|---|
| `docs/plugin_hooks.rst:1110` | `https://www.starlette.io/middleware/` | `https://starlette.dev/middleware/` |
| `docs/changelog.rst` (×3) | `https://www.uvicorn.org/` | `https://uvicorn.dev/` |
| `docs/changelog.rst:1144` | `https://hub.docker.com/datasetteproject/datasette` | `https://hub.docker.com/r/datasetteproject/datasette` |

Docs-only; no code, no behaviour change.

## Test plan

Every replacement URL was fetched and returns HTTP 200:

```
200 -> https://starlette.dev/middleware/
200 -> https://uvicorn.dev/
200 -> https://hub.docker.com/r/datasetteproject/datasette
```

The changelog entries are historical records, so the surrounding wording is untouched — only the href targets change, and they still point at the same projects' current documentation.

## Not included here

The same scan flagged other non-resolving links that I deliberately left alone, because they need a maintainer decision rather than a mechanical fix — several `*.datasettes.com` demo instances (`fivethirtyeight`, `fara`, `congress-legislators`, `san-francisco`) and some `latest.datasette.io/fixtures/...` example URLs no longer respond as documented. Happy to open a separate issue listing those with their status codes if that would be useful.


simonw:
Thanks for these.

PR #2913, opened 2026-09-15
Fix count all links to preserve filter parameters and SQLite types

simonw:
Clicking “count all” on a filtered table could report zero while the table still showed matching rows. The count-query link omitted the filter parameters, and numeric parameters supplied by plugins lost their SQLite types when passed through a URL.

Include the parameters in the count link and preserve their SQLite types. Numeric casts use unary plus to preserve the original comparison behavior; the code comment explains why both are needed. SQL literals, quoted identifiers, and comments remain intact. Query URLs also preserve the configured base URL and response format.

Validation:
- Red/green regression tests: five filtered-count cases failed before the fix and passed afterward.
- Broader table/API/utilities suite: 471 passed, 8 xpassed.
- Final focused checks: 24 passed; formatting and lint passed.
- Verified the original demo filters in the browser: “count all” returns 14,953 rows.


simonw:
I don't like the complexity of this. I'm going to go a different way:
- https://github.com/simonw/datasette/issues/2914#issuecomment-5684085237

PR #2929, opened 2026-09-16
Fix facet selection for explicit exact filters

### LINKED ISSUE #1695: Option to un-filter facet not shown for `?col__exact=value`
Spotted this on a page with `COUNTY__exact=Lee` in the URL:

CleanShot 2022-03-29 at 21 41 46@2x

With `COUNTY=Lee` you get this instead:

image

simonw:
`colname__exact=value` is now treated the same as `colname=value` when determining which facets are selected.

<details><summary>GPT-6 Astra Extra High</summary>

Filters such as `COUNTY__exact=Lee` filtered the table but left the facet value marked as unselected, hiding its × link. When both `COUNTY=Lee` and `COUNTY__exact=Lee` were present, the removal link left the explicit filter behind.

Recognize both forms when selecting a facet value and remove all matching forms when toggled off. Preserve unrelated filters, other values for the same column, and existing handling of underscore-prefixed columns. Includes a changelog entry.

Closes #1695.

Validation:
- Red: added regression tests before changing the implementation; **6 failed, 5 passed**, reproducing the missing × link and incomplete filter removal.
- Green: **all 11 regression cases pass**, including following the HTML removal link to verify the resulting rows.
- Full facet, table HTML, and filter suites: **150 passed**.
- Black, Ruff, full repository spelling checks, Cog, blacken-docs, and `git diff --check` passed.
</details>


simonw:
Now that it's deployed here's an example: https://latest.datasette.io/fixtures/facetable?_sort=pk&_facet=planet_int&planet_int__exact=1

Before the fix:

CleanShot 2026-09-16 at 2 59 33 PM@2x

After:

CleanShot 2026-09-16 at 3 02 40 PM@2x

PR #2949, opened 2026-09-19
Ignore unrelated query parameters when loading facets

pengboyu-dev:
Requests such as `/fixtures/facetable.json?_facet=state&_facets=ignored` return HTTP 500: `load_facet_configs()` accepts every key starting with `_facet`, but only assigns a facet type for `_facet` or `_facet_...`.

As a result, `_facets` reuses the previous parameter's facet type and tries to facet on a column named `ignored`. Putting `_facets` first instead raises `UnboundLocalError` when there is no metadata configuration to initialize the local variable. Metadata-defined facets can also leak their type into the unrelated parameter.

Only parse `_facet` and `_facet_...` keys as facet configuration. This preserves custom facet types while ignoring unrelated parameters instead of turning them into additional facets.

Tests cover both query-string orders, the parameter by itself, metadata-defined facets, and actual JSON table requests. All eight new cases fail before the fix and pass afterward.

Validation on Python 3.14.7 / macOS:
- `uv run pytest tests/test_facets.py tests/test_filters.py -q`: 58 passed.
- `uv run pytest -n 4 -m 'not serial' -q`: 2647 passed, 54 skipped, 6 xfailed, 15 xpassed; 143 subtests passed.
- `NO_PROXY=localhost,127.0.0.1,::1 no_proxy=localhost,127.0.0.1,::1 uv run pytest -m serial -q`: 40 passed. The initial run without the local proxy exclusion had four localhost server-test failures; the exclusion is limited to the test command.
- `uv run black --check .` and `uv run ruff check datasette tests`: passed.


simonw:
https://latest.datasette.io/fixtures/facetable?_facets=x prior to this fix returns a 500:

> `cannot access local variable 'type' where it is not associated with a value`

